PR #12, opened 2022-03-10
fix(router): compare the number of columns when adding sharding rule

agneum:
Fix the bug on comparing the length of columns slice when adding sharding rule for proxy router


x4m:
Thanks!

PR #15, opened 2022-04-07
fix(router): run world and coordinator apps

agneum:
- proxy queries that do not match any shard to the world storage
- send DDL queries to all shards
- update proto-files and open ports to run coordinator
- fix typos


x4m (approved):

Denchick:
It is a pity we did not find the time and effort to review and merge this PR. Come back with new prs, we won't let you down

PR #17, opened 2022-04-20
pg_query instead of sqlparser[WIP]

reshke:


NikolayS:
What are the pros and cons of using pg_query? Obviously, it gives better coverage in terms of PG syntax, and it's well supported (and this aspect is really important), but pg_query_go uses cgo which has certain issues (such as discussed in https://relistan.com/cgo-when-and-when-not-to-use-it).

How do we approach the decision here?

cc @x4m @agneum

PR #34, opened 2022-05-26
Make specifying the tls config optional

Denchick:


pkg/config/tls.go
```diff
@@ -18,16 +18,21 @@ type TLSConfig struct {
...
+}
+
+func (c *TLSConfig) Init() (*tls.Config, error) {
+	tracelog.InfoLogger.Printf("Init TLS kek")
+	if c.IsTLSModeDisable() {
 		tracelog.InfoLogger.Printf("skip loading tls certs")
+		return nil, nil
+	}
+
+	tracelog.InfoLogger.Printf("loading tls cert file %s, key file %s", c.CertFile, c.KeyFile)
```
reshke, line 32: sqprlog here and above

Denchick: well, using spqrlog here causes import cycle because spqrlog requires config.
I suggest getting log level not from config but from environment variable or pass log level to logger and every service should create its own logger with its own level from its own config. 
But maybe fix this next time ¯\_(ツ)_/¯

reshke (approved):

PR #38, opened 2022-05-31
TLS and rrouter small refactoring

Denchick:
I just tried to make the shard configs work, but it didn't work.
- Removed unnecessary passing of `sslmode` to different functions. If `*tls.Config` is not nil, then ssl is enabled.
- Fixed those code with initializing frontend and backend rules


pkg/config/tls.go
```diff
@@ -7,24 +7,18 @@ import (
...
 type TLSConfig struct {
 	SslMode  string `json:"sslmode" toml:"sslmode" yaml:"sslmode"`
 	KeyFile  string `json:"key_file" toml:"key_file" yaml:"key_file"`
 	CertFile string `json:"cert_file" toml:"cert_file" yaml:"cert_file"`
 }
 
-func (c *TLSConfig) IsTLSModeDisable() bool {
-	return c == nil || c.SslMode == SSLMODEDISABLE || (c.CertFile == "" && c.KeyFile == "" || c.SslMode == "")
+func (c *TLSConfig) IsSSLModeRequired() bool {
+	return c != nil && c.SslMode == "require"
```
reshke, line 17: why do we drop SSLMODEDISABLE constant here?
how about sslmode='verify-full'?

PR #106, opened 2023-01-13
Allow dataspace create & list in local proxy router setup

reshke:


Denchick:
This pr didn't pass tests, so I revert it

PR #124, opened 2023-01-17
Sharding rules as table

diPhantxm:


test/regress/tests/console/expected/add.out
```diff
@@ -88,9 +88,9 @@ SHOW key_ranges;
 (1 row)
 
 SHOW sharding_rules;
-                     listing sharding rules                      
------------------------------------------------------------------
- sharding rule rule1 for table (*) with columns [id, hash: x->x]
+ Sharding Rule ID | Table Name | Columns 
+------------------+------------+---------
+ rule1            | *          | id
```
reshke, line 93: Also, hash function used for column should appear in the output

Denchick: Although there are no implementations of functions except `x->x`

PR #126, opened 2023-01-17
Move lambda function call from proxy internals to

reshke:
auth package

### DIFF
--- a/pkg/conn/auth.go
+++ b/pkg/conn/auth.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"fmt"
 
+	"github.com/pg-sharding/spqr/pkg/client"
 	"github.com/pg-sharding/spqr/pkg/spqrlog"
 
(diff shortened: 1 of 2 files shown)


router/pkg/rule/rulemgr.go
```diff
@@ -94,7 +102,7 @@ func (m *MgrImpl[T]) MatchKey(key route.Key) (*T, error) {
 	}
 
 	return nil, fmt.Errorf("failed to route frontend for client:"+
-		" route for user:%s and db:%s is unconfigured", key.Usr(), key.DB())
+		" route for user:%s and db:%s is unconfigured in %s", key.Usr(), key.DB(), m.UnderlyingEntityName())
```
reshke, line 105: Useful for debug
```
psql: error: connection to server at "localhost" (::1), port 6432 failed: failed to route frontend for client: route for user:usermd5 and db:db1 is unconfigured in backend rules
```

PR #139, opened 2023-05-02
Accept and parse comments in queries to force-set shard to ruote

reshke:
Comment format expected tio be:
key: value[, key1: value1...]


x4m (approved):
Looks great!

PR #144, opened 2023-05-30
Route client to admin console on port psql port if dbname is "console"

reshke:


router/pkg/rulerouter/rulerouter.go
```diff
@@ -137,6 +137,10 @@ func (r *RuleRouterImpl) PreRoute(conn net.Conn) (rclient.RouterClient, error) {
 		return cl, err
 	}
 
+	if cl.DB() == "console" {
```
Denchick, line 140: Maybe we should use a less common name? Like spqr-console or something?

PR #150, opened 2023-06-07
 Support cancel proto msgs 

reshke:


router/pkg/rulerouter/rulerouter.go
```diff
@@ -249,4 +266,33 @@ func (r *RuleRouterImpl) Config() *config.Router {
...
+	r.clmp[cl.GetCancelPid()] = cl
+}
+
+func (r *RuleRouterImpl) ReleaseClient(cl rclient.RouterClient) {
+	r.clmu.Lock()
+	defer r.clmu.Unlock()
+	delete(r.clmp, cl.GetCancelPid())
+}
+
+func (r *RuleRouterImpl) CancelCleint(csm *pgproto3.CancelRequest) error {
```
Denchick, line 281: CancelClient

router/pkg/rulerouter/rulerouter.go
```diff
@@ -249,4 +266,33 @@ func (r *RuleRouterImpl) Config() *config.Router {
...
+		if cl.GetCancelKey() != csm.SecretKey {
+			return fmt.Errorf("cancel sekret does not match")
+		}
+		if cl.Server() != nil {
+			spqrlog.Logger.Printf(spqrlog.DEBUG1, "cancelling client pid %d", csm.ProcessID)
+			return cl.Server().Cancel()
+		}
+		return nil
+	}
+	return fmt.Errorf("no client with pid %d", csm.ProcessID)
```
Denchick, line 295: Im not sure if we should return an error here. Is this error will be passing to the client?
Seems like if there is no process, then client was already released?

test/regress/tests/router/expected/target_session_attrs.out
```diff
@@ -31,4 +31,3 @@ SELECT pg_is_in_recovery() /* target-session-attrs: read-only */ , id FROM tsa_t
 NOTICE: send query to shard(s) : sh2
 ERROR:  failed to find replica
 message type 0x5a arrived from server while idle
-ERROR:  failed to find replica
```
Denchick, line 34: Why are you drop this changes?

router/pkg/rulerouter/rulerouter.go
```diff
@@ -32,6 +33,13 @@ type RuleRouter interface {
 	AddWorldShard(key qdb.ShardKey) error
 	AddShardInstance(key qdb.ShardKey, host string)
 
+	CancelCleint(csm *pgproto3.CancelRequest) error
+	AddClient(
+		cl rclient.RouterClient)
```
Denchick, line 38: strange)

router/pkg/rulerouter/rulerouter.go
reshke: this error will not be passed to the client

PR #151, opened 2023-06-09
Reload logger when getting SIGHUP SIGUSR1 signals

Denchick:


reshke (approved):

router/pkg/client/client.go
```diff
@@ -543,7 +543,6 @@ func (cl *PsqlClient) Auth(rt *route.Route) error {
 
 	ps, err := rt.Params()
 	if err != nil {
-		spqrlog.Logger.PrintError(err)
```
reshke, line 546: why

PR #168, opened 2023-06-21
use qlog in memdb

reshke:


Denchick:
Close as stale due to #244

PR #184, opened 2023-06-28
Add kill client command

nikifkon:
Example of usage:

Admin console:

```
nikifkon=> show clients;
 client id | user | dbname 
-----------+------+--------
(0 rows)

nikifkon=> show clients;
  client id   |   user   | dbname 
--------------+----------+--------
 0xc0004be240 | nikifkon | db1
(1 row)

nikifkon=> kill client 0xc0004be240;
            kill client             
------------------------------------
 the client 0xc0004be240 was killed
(1 row)

nikifkon=> show clients;
 client id | user | dbname 
-----------+------+--------
(0 rows)
```

Client console:
```
nikifkon@nikifkon-osx spqr % psql -h localhost -p 6432 -d db1
psql (14.8 (Homebrew))
Type "help" for help.

db1=# select 1;
backend is shutdown, your connection closed
db1=?# select 1;
no connection to the server
The connection to the server was lost. Attempting reset: Succeeded.
db1=# select 1;
NOTICE: send query to shard(s) : sh1,sh2
 ?column? 
----------
        1
        1
(2 rows)

db1=# 
```

1. pg_regress does not support random values inside expoected.out files, so I don't know how to tests this feature
2. do I need to cancel the request of the killed client?


nikifkon:
it turns out that the shards from the killed client is not synchronized, so it is automatically discarded

reshke (approved):

PR #202, opened 2023-07-10
Implement SCRAM auth

EinKrebs:


pkg/auth/auth.go
```diff
@@ -53,6 +54,61 @@ func AuthBackend(shard conn.DBInstance, berule *config.BackendRule, msg pgproto3
 			return fmt.Errorf("no auth rule specified for server connection")
 		}
 		return shard.Send(&pgproto3.PasswordMessage{Password: berule.AuthRule.Password})
```
reshke, line 56: what about authFrontend() ? 

also, we can actually setup postgresql shard with scram auth in tests and run scram auth in CI....

reshke:
Also pleasse indicate in the documentation/readme that SCRAM is now supported 🦀️️️️️️🦀️️️️️️🦀️️️️️️

EinKrebs:
> Also pleasse indicate in the documentation/readme that SCRAM is now supported crab️️️️️crab️️️️️crab️️️️️

Where to write this? Do I need to create another .md file in docs/ or update existing?

Denchick:
> Where to write this? Do I need to create another .md file in docs/ or update existing?

I would suggest create a new section Authentication methods or something in Router.md file.

reshke:
So, basically scram test are the same as non-scram test with exception in auth process be each sql-file run? Maybe we can add unit test for this feature too? See https://github.com/pg-sharding/spqr/blob/master/test/authentification/auth_test.go

EinKrebs:
> So, basically scram test are the same as non-scram test with exception in auth process be each sql-file run? Maybe we can add unit test for this feature too? See https://github.com/pg-sharding/spqr/blob/master/test/authentification/auth_test.go

It seems rather hard to mock psql's behavior in unit tests, it's way easier to set up cluster in docker compose

reshke (approved):

Denchick (changes requested):
Let's not test SCRUM in docker

Denchick (approved):

PR #205, opened 2023-07-11
Enable different passwords for shards

debebantur:


reshke:
What about tests?

debebantur:
> What about tests?

Added unit tests

reshke (approved):

PR #212, opened 2023-07-11
Hide router notice msgs under extended debug setting

reshke:


pkg/config/router.go
```diff
@@ -33,6 +33,7 @@ type Router struct {
 
 	MaintainParams bool `json:"maintain_params" toml:"maintain_params" yaml:"maintain_params"`
 	WithJaeger     bool `json:"with_jaeger" toml:"with_jaeger" yaml:"with_jaeger"`
+	ExtendedDebug  bool `json:"extended_debug" toml:"extended_debug" yaml:"extended_debug"`
```
Denchick, line 36: Why not just debug?

router/rulerouter/relay.go
```diff
@@ -63,6 +65,8 @@ type RelayStateImpl struct {
 	WorldShardFallback bool
 	routerMode         config.RouterMode
 
+	extendedDebug bool
```
Denchick, line 68: Why you just don't getting the setting from router's config?

router/rulerouter/relay.go
```diff
@@ -186,9 +195,11 @@ func (rst *RelayStateImpl) procRoutes(routes []*qrouter.DataShardRoute) error {
 	for _, shr := range routes {
 		rst.activeShards = append(rst.activeShards, shr.Shkey)
 	}
-	// TDB: hide under setting
-	if err := rst.Cl.ReplyDebugNoticef("matched datashard routes %+v", routes); err != nil {
-		return err
+	if rst.ExtendedDebug() {
+		// TDB: hide under setting
```
Denchick, line 199: remove TODO then)

x4m (approved):

router/rulerouter/relay.go
reshke: > Why you just don't get the setting from router's config?

because of reload

pkg/config/router.go
reshke: renamed

PR #243, opened 2023-07-24
Show clients on coordinator as well

nikifkon:
example output:

```
postgres=> show clients;
  client_id   |   user   |  dbname  |       server_id       | router_address 
--------------+----------+----------+-----------------------+----------------
 0xc0003ffad0 | postgres | postgres | no backend connection | localhost:7001
 0xc0000f09c0 | postgres | postgres | localhost:5001        | localhost:7000
 0xc0000f09c0 | postgres | postgres | localhost:5002        | localhost:7000
 0xc0003fef70 | postgres | postgres | no backend connection | localhost:7000
(4 rows)
```


coordinator/provider/coordinator.go
```diff
@@ -32,6 +35,139 @@ import (
 	spqrparser "github.com/pg-sharding/spqr/yacc/console"
 )
 
+type FakeClient struct {
```
reshke, line 38: can we just move FakeClient from router/client to pkg/client and reuse it?

pkg/client/clientpool.go
```diff
@@ -6,8 +6,23 @@ import (
 	spqrlog "github.com/pg-sharding/spqr/pkg/spqrlog"
 )
 
+type RouterClient interface {
```
reshke, line 9: Can we choose a better name? there is already struct with this exact name

reshke (changes requested):

coordinator/provider/coordinator.go
nikifkon: Nice thought. Done. I also rename it to MockClient

pkg/client/clientpool.go
nikifkon: I think `ClientInfo` will suit, because it correspond to the same structure as in protobuf

reshke (approved):

PR #247, opened 2023-07-26
Fix release job

EinKrebs:
Replace version number in debian/control correctly

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,15 +27,18 @@ jobs:
               dep ensure
           fi
           sudo apt update && sudo apt install -y debhelper
+      
+      - name: Set env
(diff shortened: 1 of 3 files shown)


reshke (approved):

Denchick:
Before merging this PR, we should somehow test the release action. I didn't find out how to run the action manually. Maybe one of you will do it. Should we temporarily allow to run release action for every commit to ensure it is working now and then roll back it again?

EinKrebs:
> Before merging this PR, we should somehow test the release action. I didn't find out how to run the action manually. Maybe one of you will do it. Should we temporarily allow to run release action for every commit to ensure it is working now and then roll back it again?

I ran this job on my fork, it works fine

PR #250, opened 2023-07-28
test writing to qdb

debebantur:
I thought through some ways of implementing fault injection, however minor logic change seemed more reasonable


coordinator/provider/coordinator.go
```diff
@@ -237,10 +237,10 @@ func NewCoordinator(db qdb.QDB) *qdbCoordinator {
 
 	for _, r := range ranges {
 		tx, err := db.GetTransferTx(context.TODO(), r.KeyRangeID)
-		if tx == nil || err != nil || tx.ToStatus == "" {
+		if err != nil {
 			continue
 		}
-		if tx.ToStatus == "commit" {
+		if tx.ToStatus == "commit" && tx.FromStatus != "commit" {
```
reshke, line 243: maybe `tx.ToStatus/tx.FromStatus` should be enum, not string?

pkg/datatransfers/data_transfers.go
```diff
@@ -138,21 +138,21 @@ func beginTransactions(ctx context.Context, f, t string) error {
 }
 
 func commitTransactions(ctx context.Context, f, t string, krid string, db *qdb.QDB) error {
```
reshke, line 140: maybe we can tests this with unit tests also? 
We can create mock - implementation of qdb, which will return expected results on all operations

pkg/config/data_transfers.go
```diff
@@ -41,13 +41,12 @@ func LoadShardDataCfg(cfgPath string) (*DatatransferConnections, error) {
...
 
-	configBytes, err := json.MarshalIndent(cfg, "", "  ")
+	_, err = json.MarshalIndent(cfg, "", "  ")
 	if err != nil {
 		return &cfg, err
 	}
 
-	fmt.Println("Running config:", string(configBytes))
-
+	//log.Println("Running config:", string(configBytes))
```
reshke, line 49: do not comment code, remove it

pkg/datatransfers/data_transfers.go
```diff
@@ -62,25 +78,36 @@ func MoveKeys(ctx context.Context, fromId, toId string, keyr qdb.KeyRange, shr [
 		}
 	}
 
-	err := beginTransactions(ctx, fromId, toId)
+	from, err := pgx.Connect(ctx, createConnString(fromId))
```
reshke, line 81: maybe we need brief comment on MoveKeys function? 

something like: 
/*
MoveKeys: performs physical key-range move from one datashard to another. 
Assumes that passed key range already locked on every online sqpr-router. 
Steps:
1) Traverse pg_class to resolve all relations that matches some sharding rules 
2) create sql `copy` and `delete` statements to move data tuples.
3) prepare and commit distributed move transation 
*/

pkg/datatransfers/data_transfers_test.go
reshke, line 1: nice

pkg/datatransfers/data_transfers.go
```diff
@@ -27,22 +28,37 @@ type ProxyW struct {
...
 	return p.w.Write(bt)
 }
 
 var shards *config.DatatransferConnections
-var txFrom pgx.Tx
-var txTo pgx.Tx
-var localConfigDir = "/pkg/datatransfers/shard_data.yaml"
+var lock sync.RWMutex
+
+var localConfigDir = "/shard_data.yaml"
```
reshke, line 44: this should be somewhere else. like `cmd/mover`

pkg/mock/mock_pgxconn_ifase.go
```diff
@@ -0,0 +1,80 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: ../Documents/spqr/pkg/datatransfers/temp.go
```
reshke, line 2: where is `datatransfers/temp.go` file itself? It should not be dropped. 
Also, we need Makefile target which runs `go generate .... ` on these files

reshke (approved):

PR #253, opened 2023-08-01
Coordinator tests

diPhantxm:
intervals are now half intervals [0, 10)
fixed lock/unlock, split/unite functions in coordinator
coordinator can restart now
if router is down, coordinator just skips it in traverse
if qdb is unavailable, coordinator is not down, but cannot perform any operations


coordinator/provider/coordinator.go
```diff
@@ -506,20 +549,44 @@ func (qc *qdbCoordinator) Split(ctx context.Context, req *kr.SplitKeyRange) erro
 		},
 	)
 
+	// splitting by X makes X the point of intersection
```
reshke, line 552: is this comment actual?

test/feature/features/coordinator.feature
```diff
@@ -0,0 +1,235 @@
...
+    SHOW routers
+    """
+    Then SQL result should match regexp
+    """
+    \[\]
+    """
+
+    When I run SQL on host "coordinator"
+    """
+    REGISTER ROUTER r2 ADDRESS regress_router::7000;
```
Denchick, line 42: What happen if you register two routers with the same address?

test/feature/features/coordinator.feature
```diff
@@ -0,0 +1,235 @@
...
+  Scenario: QDB is down
+    Given host "qdb01" is stopped
+    When I run SQL on host "coordinator"
+    """
+    CREATE KEY RANGE krid3 FROM 31 TO 40 ROUTE to sh1;
+    """
+    Then command return code should be "1"
+    And SQL error on host "coordinator" should match regexp
+    """
+    context deadline exceeded
```
Denchick, line 209: Maybe we should explicitly say that QDB is not available?

Denchick (changes requested):

pkg/clientinteractor/interactor.go
reshke, line 1: Nice

test/feature/features/coordinator.feature
diPhantxm: Yeah, I agree, because `context deadline exceeded` is the error we get because we wait for query response for too long. It happens in many different ways, for instance, when something is locked (key range) or down (qdb).
Now goroutine, processing the query, just hangs until that something is unlocked or up again. I thought maybe we should check if required object is locked or down and if so return an error with appropriate message. 
What you think? I can try to do it in the next pr.

test/feature/features/coordinator.feature
diPhantxm: I'll extend the scenario and push commit

test/feature/features/coordinator.feature
```diff
@@ -48,6 +48,45 @@ Feature: Coordinator test
     router r2-regress_router:7000
     """
 
+  Scenario: Register 2 routers with same address
```
Denchick, line 51: Register two routers with same address fails

test/feature/features/coordinator.feature
Denchick: > maybe we should check if required object is locked or down

agree

> I can try to do it in the next pr.

lets do this

Denchick (changes requested):

coordinator/provider/coordinator.go
```diff
@@ -432,7 +442,12 @@ func (qc *qdbCoordinator) AddKeyRange(ctx context.Context, keyRange *kr.KeyRange
 		})
 
 		if err != nil {
-			return err
+			spqrlog.Zero.Debug().
+				Str("router", r.ID).
+				Err(err).
+				Msg("etcdqdb: notify router add key range")
```
Denchick, line 448: nice but only without etcdqdb

coordinator/provider/coordinator.go
```diff
@@ -432,7 +442,12 @@ func (qc *qdbCoordinator) AddKeyRange(ctx context.Context, keyRange *kr.KeyRange
...
+			continue
+			//return err
```
Denchick, line 450: why?

router/client/client.go
```diff
@@ -571,17 +571,14 @@ func (cl *PsqlClient) Init(tlsconfig *tls.Config) error {
...
-				return err
-			}
 		case conn.CANCELREQ:
 			cl.csm = &pgproto3.CancelRequest{}
 			if err = cl.csm.Decode(msg); err != nil {
 				return err
 			}
 
-			return nil
+			//return nil
```
Denchick, line 580: Why?

test/feature/features/coordinator.feature
```diff
@@ -0,0 +1,279 @@
...
+    When I run SQL on host "coordinator"
+    """
+    REGISTER ROUTER r3 ADDRESS invalid_router::7000
+    """
+    Then SQL error on host "coordinator" should match regexp
+    """
+    Error while dialing
+    """
+
+  Scenario: Unregister router with invalid id fails
```
Denchick, line 99: Seem like it works

test/regress/tests/router/expected/mixed_routing.out
```diff
@@ -204,12 +204,12 @@ NOTICE: send query to shard(s) : sh1
 ----
 (0 rows)
 
-SELECT * FROM xxmixeduuid WHERE id <= 'FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF' ORDER BY id;
+SELECT * FROM xxmixeduuid WHERE id <= 'FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFE' ORDER BY id;
```
Denchick, line 207: Maybe rollback all changes in test/regress/tests/router (except this file)? Because this changes seem unneeded

test/feature/features/coordinator.feature
Denchick: btw nice tests!

test/regress/tests/router/expected/mixed_routing.out
Denchick: And router/qrouter/proxy_routing_test.go

Denchick (changes requested):

coordinator/provider/coordinator.go
diPhantxm: I was not sure here. As I see the code, we could use `traverseRouters` to notify about new key ranges, and in that function we notify all the routers even if error occurred somewhere. But here we wanted to stop if any error raised

test/regress/tests/router/expected/mixed_routing.out
diPhantxm: Well I didn't really check what tests need correction, I just wanted to keep test logic unchanged

router/client/client.go
diPhantxm: I don't remember exactly, probably it crashed here. I checked it rn, seems like it's working, restored changes. Will note down such things next time

router/console/console.go
```diff
@@ -84,9 +84,18 @@ https://github.com/pg-sharding/spqr/tree/master/docs
...
+	for _, p := range params {
+		if v, ok := cl.Params()[p]; ok {
+			msgs = append(msgs, &pgproto3.ParameterStatus{Name: p, Value: v})
+		}
+	}
+
+	msgs = append(msgs, []pgproto3.BackendMessage{
 		&pgproto3.ParameterStatus{Name: "integer_datetimes", Value: "on"},
 		&pgproto3.ParameterStatus{Name: "server_version", Value: "console"},
 		&pgproto3.ParameterStatus{Name: "standard_conforming_strings", Value: "on"},
```
nikifkon, line 101: can be deleted with new changes 🙃

Denchick (approved):

PR #254, opened 2023-08-01
Statistics fix

debebantur:


router/statistics/query_time_statistics.go
```diff
@@ -25,12 +26,14 @@ type statistics struct {
 	TimeData          map[string]*startTimes
 	Quantiles         []float64
 	NeedToCollectData bool
+	lock              sync.RWMutex
```
Denchick, line 29: Maybe you actually need [sync.Map](https://pkg.go.dev/sync#Map)

debebantur: I read about it, it seems that sync.Map is optimal to use when we mostly read from map and have many cores. Is it the case?

router/statistics/query_time_statistics.go
```diff
@@ -81,16 +90,21 @@ func RecordFinishedTransaction(t time.Time, client string) {
 	if err != nil {
 		spqrlog.Zero.Error().Err(err).Msg(err.Error())
 	}
+	queryStatistics.lock.Unlock()
 }
 
 func GetClientTimeStatistics(tip StatisticsType, client string) *tdigest.TDigest {
 	var stat *tdigest.TDigest
+
+	queryStatistics.lock.Lock()
```
reshke, line 99: defer queryStatistics.lock.Unlock()

Denchick (approved):

reshke (approved):

PR #258, opened 2023-08-03
Make `--console` actually work. Check for simultaneous use of console…

reshke:
…/daemonize

### DIFF
--- a/cmd/router/main.go
+++ b/cmd/router/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"log"
 	_ "net/http/pprof"
 	"os"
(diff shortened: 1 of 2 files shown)


cmd/router/main.go
```diff
@@ -81,7 +82,11 @@ var runCmd = &cobra.Command{
 			return err
 		}
 
-		if rcfg.Daemonize || daemonize {
+		if console && daemonize {
```
Denchick, line 85: Why not (rcfg.Daemonize || daemonize) ?

reshke: Because cli options are stronger than config pptions

Denchick (approved):

PR #262, opened 2023-08-03
fix init.sql + backup.json logic

nikifkon:
previously `init.sql` apply if backup file **exists** rather then not exist


docker/router/Dockerfile
```diff
@@ -1,3 +1,3 @@
 FROM spqr-base-image
 
-ENTRYPOINT /spqr/spqr-router run -c ${SPQR_CONFIG=/spqr/docker/router/cfg.yaml} --proto-debug
+ENTRYPOINT /spqr/spqr-router run -c ${ROUTER_CONFIG=/spqr/docker/router/cfg.yaml} --proto-debug
```
Denchick, line 3: Nice

docs/Router.md
```diff
@@ -21,22 +21,22 @@ All SPQR configurations can be written in json, yaml or toml format. See example
...
+| `router_port`          | the router port                                                                                                                                                                                             |
+| `admin_console_port`   | the admin console port                                                                                                                                                                                      |
+| `grpc_api_port`        | the API port                                                                                                                                                                                                |
+|                        |                                                                                                                                                                                                             |
+| `init_sql`             | a path to a sql commands, that will be run on the startup of the router (if no file was found along the path `memqdb_backup_path`)                                                                          |
```
Denchick, line 33: No need in brackets. Just

> a path to a SQL command, that will be run on the router's startup. It will be ignored if memqdb_backup_path exists.

docs/Router.md
```diff
@@ -21,22 +21,22 @@ All SPQR configurations can be written in json, yaml or toml format. See example
...
+| `jaeger_url`           | a path to the Jaeger instance - open source software for tracing transactions between distributed services                                                                                                  |
+| `world_shard_fallback` | can be true or false. If false, then router will raise an error when query will be impossible to send to particular shard. If true, then router will route unrouted query to the world shard.               |
+| `show_notice_messages` | can be true or false. May help to debug the router and see to where it is actualy sending queries                                                                                                           |
+| `time_quantiles`       | list of time quantiles to show querry time statistics. When empty, no statistics is collected                                                                                                               |
+| `memqdb_backup_path`   | if specified, memqdb will backup state into a file. At the start of the Router, if a file with a backup is exist, then the state of the MemQDB will be restored, otherwise init.sql will be used (if any)   |
```
Denchick, line 39: > MemQDB backup state path. MemQDB's state restored if a file backup exists during the router startup. If there is no file, init.sql will be used.

test/feature/features/memqdb.feature
```diff
@@ -60,7 +88,81 @@ Feature: MemQDB save state into a file
...
+          "Table Name":"test"
+      },
+      {
+          "Columns":"idy",
+          "Hash Function":"x->x",
+          "Sharding Rule ID":"rule3",
+          "Table Name":"*"
+      }
+    ]
+    """
```
Denchick, line 159: It would be cool to modify rules via spqr-console here, restart router and ensure that backup file works

Denchick: And what happen if backup file and Init SQL is specified but files does not exists?

Denchick (changes requested):

test/feature/features/memqdb.feature
nikifkon: > It would be cool to modify rules via spqr-console here, restart router and ensure that backup file works

I did it at `Scenario: Sharding rules restored after droping specific one (init.sql ignored)`

nikifkon: > And what happen if backup file and Init SQL is specified but files does not exists?

I will add test with `Scenario: Sharding rules not initialized if `init.sql` file doesn't exists`

Backup file don't exists in all scenarios, so this case already tested

docs/Router.md
nikifkon: thanks

test/feature/features/memqdb.feature
nikifkon: `Scenario: Sharding rules not initialized if init.sql file doesn't exists`

it turned out that router failed to start in this scenario. Maybe it make scene?

Denchick (approved):

PR #265, opened 2023-08-04
Coordinator Show clients/pools/backend_connections funcs

debebantur:
Created SHOW pools in coordinator and added feature tests for SHOW pools/clients/backend_connections


coordinator/provider/coord_pool.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+		CoordPoolIterator: CoordPoolIterator{},
+	}
+}
+
+func (r *CoordPool) Connection(clid string, shardKey kr.ShardKey) (shard.Shard, error) {
+	return nil, fmt.Errorf("unimplemented")
+}
+
+type CoordConnectionKepper struct {
```
reshke, line 38: can we reuse some generic structure from `pkg`?

test/feature/conf/router2.yaml
```diff
@@ -0,0 +1,44 @@
+host: 'regress_router_2'
```
reshke, line 1: the whole reason to define second config to redefine `host`?

debebantur: yep, maybe it is possible to make it with environmental vars, however i am confused how to do it

reshke (approved):

PR #271, opened 2023-08-10
fix double lock key range on router when moving data

diPhantxm:
lock and unlock only once in Move
and change method name in interface to be more consistent

### DIFF
--- a/coordinator/provider/coordinator.go
+++ b/coordinator/provider/coordinator.go
@@ -531,7 +531,7 @@ func (qc *qdbCoordinator) LockKeyRange(ctx context.Context, keyRangeID string) (
 	})
 }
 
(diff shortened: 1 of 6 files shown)


coordinator/provider/coordinator.go
```diff
@@ -531,7 +531,7 @@ func (qc *qdbCoordinator) LockKeyRange(ctx context.Context, keyRangeID string) (
 	})
 }
 
-func (qc *qdbCoordinator) Unlock(ctx context.Context, keyRangeID string) error {
+func (qc *qdbCoordinator) UnlockKeyRange(ctx context.Context, keyRangeID string) error {
```
Denchick, line 534: 👍

Denchick (approved):
LGTM

PR #275, opened 2023-08-11
add some checks to Split/Unite

diPhantxm:
Unite:
* check shardID is the same on left and right key ranges
* check the UpperBound on the left key range matches LowerBound on the right key range

Split:
* check that bound is not less than lower bound
* check that bound is not greater than or equal to upper bound

### DIFF
--- a/coordinator/provider/coordinator.go
+++ b/coordinator/provider/coordinator.go
@@ -578,6 +578,10 @@ func (qc *qdbCoordinator) Split(ctx context.Context, req *kr.SplitKeyRange) erro
 		}
 	}()
 
+	if kr.CmpRangesLess(req.Bound, krOld.LowerBound) || !kr.CmpRangesLess(req.Bound, krOld.UpperBound) {
+		return fmt.Errorf("bound is out of key range")
+	}
+
 	krNew := kr.KeyRangeFromDB(
 		&qdb.KeyRange{
 			LowerBound: req.Bound,
--- a/coordinator/provider/coordinator.go
+++ b/coordinator/provider/coordinator.go
@@ -705,6 +705,16 @@ func (qc *qdbCoordinator) Unite(ctx context.Context, uniteKeyRange *kr.UniteKeyR
 		}
 	}()
 
+	if krLeft.ShardID != krRight.ShardID {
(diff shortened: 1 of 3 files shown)


test/feature/spqr_test.go
```diff
@@ -42,7 +42,7 @@ const (
 	consoleName                     = "spqr-console"
 	postgresqlConnectTimeout        = 30 * time.Second
 	postgresqlInitialConnectTimeout = 2 * time.Minute
-	postgresqlQueryTimeout          = 5 * time.Second
+	postgresqlQueryTimeout          = 7 * time.Second
```
Denchick, line 45: Maybe 10 then? 😄

coordinator/provider/coordinator.go
```diff
@@ -705,6 +709,16 @@ func (qc *qdbCoordinator) Unite(ctx context.Context, uniteKeyRange *kr.UniteKeyR
...
 	}()
 
+	if krLeft.ShardID != krRight.ShardID {
+		return fmt.Errorf("failed to unite key ranges routing different shards")
+	}
+	if !kr.CmpRangesEqual(krLeft.UpperBound, krRight.LowerBound) {
+		if !kr.CmpRangesEqual(krLeft.LowerBound, krRight.UpperBound) {
+			return fmt.Errorf("failed to unite not adjacent key ranges")
+		}
+		return fmt.Errorf("failed to unite key ranges in wrong order")
```
Denchick, line 719: Maybe you should do it somehow?

Denchick (changes requested):

test/feature/features/coordinator.feature
```diff
@@ -273,6 +273,76 @@ Feature: Coordinator test
...
+
+  Scenario: Unite key ranges routing different shards fails
+    When I run SQL on host "coordinator"
+    """
+    CREATE KEY RANGE krid3 FROM 31 TO 40 ROUTE TO sh1;
+    UNITE KEY RANGE krid2 WITH krid3
+    """
+    Then SQL error on host "coordinator" should match regexp
+    """
+    failed to unite key ranges routing different shards
```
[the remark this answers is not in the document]
diPhantxm: Why you think so?

Denchick (approved):

PR #276, opened 2023-08-14
fix key range intersection checks

diPhantxm:
* added feature tests for adding intersecting key range
* fix adding existing krid by splitting
* fix moving key range to the same shardID

If you move key range to the same shardID, it will return an error with the following message: `ERROR:  ERROR: transaction identifier "sh2-krid4" is already in use (SQLSTATE 42710)`. So I thought it's kinda pointless to move to the same shard and added another check.

Also when adding a new key range, one case of ranges intersection was uncovered - when you add a key range that's inside another existing key range. Thus this key range is added, but they actually intersect.

### DIFF
--- a/coordinator/provider/coordinator.go
+++ b/coordinator/provider/coordinator.go
@@ -812,6 +812,11 @@ func (qc *qdbCoordinator) Move(ctx context.Context, req *kr.MoveKeyRange) error 
 	//move between shards
 	keyRange, _ := qc.db.GetKeyRange(ctx, req.Krid)
 	shardingRules, _ := qc.ListShardingRules(ctx)
+
+	if keyRange.ShardID == req.ShardId {
+		return nil
+	}
+
 	/* physical changes on shards */
 	err = datatransfers.MoveKeys(ctx, keyRange.ShardID, req.ShardId, *keyRange, shardingRules, qc.db)
 	if err != nil {
--- a/coordinator/provider/coordinator.go
+++ b/coordinator/provider/coordinator.go
@@ -579,6 +579,9 @@ func (qc *qdbCoordinator) Split(ctx context.Context, req *kr.SplitKeyRange) erro
 		}
 	}()
 
+	if kr.CmpRangesEqual(req.Bound, krOld.LowerBound) || kr.CmpRangesEqual(req.Bound, krOld.UpperBound) {
+		return fmt.Errorf("failed to split because bound equals lower or upper bound of the key range")
+	}
 	if kr.CmpRangesLess(req.Bound, krOld.LowerBound) || !kr.CmpRangesLess(req.Bound, krOld.UpperBound) {
 		return fmt.Errorf("failed to split because bound is out of key range")
 	}
--- a/coordinator/provider/coordinator.go
+++ b/coordinator/provider/coordinator.go
@@ -580,6 +580,9 @@ func (qc *qdbCoordinator) Split(ctx context.Context, req *kr.SplitKeyRange) erro
 	if kr.CmpRangesLess(req.Bound, krOld.LowerBound) || !kr.CmpRangesLess(req.Bound, krOld.UpperBound) {
 		return fmt.Errorf("failed to split because bound is out of key range")
 	}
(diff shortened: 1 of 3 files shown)


coordinator/provider/coordinator.go
```diff
@@ -581,6 +581,9 @@ func (qc *qdbCoordinator) Split(ctx context.Context, req *kr.SplitKeyRange) erro
 	if kr.CmpRangesLess(req.Bound, krOld.LowerBound) || !kr.CmpRangesLess(req.Bound, krOld.UpperBound) {
 		return fmt.Errorf("failed to split because bound is out of key range")
 	}
+	if _, err := qc.db.GetKeyRange(ctx, req.Krid); err == nil {
```
Denchick, line 584: You could make this check even before locking key range.

coordinator/provider/coordinator.go
```diff
@@ -766,6 +769,11 @@ func (qc *qdbCoordinator) Move(ctx context.Context, req *kr.MoveKeyRange) error
 	//move between shards
 	keyRange, _ := qc.db.GetKeyRange(ctx, req.Krid)
 	shardingRules, _ := qc.ListShardingRules(ctx)
+
+	if keyRange.ShardID == req.ShardId {
+		return nil
```
Denchick, line 774: Its quite obvious but it would be nice to add some comments why

qdb/ops/ops.go
```diff
@@ -51,8 +51,9 @@ func AddKeyRangeWithChecks(ctx context.Context, qdb qdb.QDB, keyRange *kr.KeyRan
 	}
 
 	for _, v := range existsKrids {
-		if kr.CmpRangesLess(keyRange.LowerBound, v.LowerBound) && kr.CmpRangesLess(v.LowerBound, keyRange.UpperBound) ||
-			kr.CmpRangesLess(keyRange.LowerBound, v.UpperBound) && kr.CmpRangesLess(v.UpperBound, keyRange.UpperBound) {
+		if kr.CmpRangesLessEqual(keyRange.LowerBound, v.LowerBound) && kr.CmpRangesLess(v.LowerBound, keyRange.UpperBound) ||
```
Denchick, line 54: Maybe its time to fix https://github.com/pg-sharding/spqr/pull/276/files#diff-6a2b55b44547e95606a78cb12f14a8009366673e342d5e813bcf2f4ff558f1fdR39

For example, qdb.GetKeyRange method is called twice

test/feature/features/coordinator.feature
```diff
@@ -428,4 +428,50 @@ Feature: Coordinator test
...
+    """
+    Then command return code should be "0"
+
+    When I run SQL on host "coordinator"
+    """
+    ADD KEY RANGE krid4 FROM 90 TO 105 ROUTE TO sh1
+    """
+    Then SQL error on host "coordinator" should match regexp
+    """
+    key range krid4 intersects with krid3 present in qdb
```
Denchick, line 449: key range A intersects with key range B in QDB

qdb/ops/ops.go
```diff
@@ -124,8 +125,9 @@ func ModifyKeyRangeWithChecks(ctx context.Context, qdb qdb.QDB, keyRange *kr.Key
 			// update req
 			continue
 		}
-		if kr.CmpRangesLess(keyRange.LowerBound, v.LowerBound) && kr.CmpRangesLess(v.LowerBound, keyRange.UpperBound) ||
-			kr.CmpRangesLess(keyRange.LowerBound, v.UpperBound) && kr.CmpRangesLess(v.UpperBound, keyRange.UpperBound) {
+		if kr.CmpRangesLessEqual(keyRange.LowerBound, v.LowerBound) && kr.CmpRangesLess(v.LowerBound, keyRange.UpperBound) ||
```
Denchick, line 128: It would be cool to add some ASCII art here

Denchick (changes requested):

diPhantxm:
When splitting, check that bound doesn't equal lower or upper bound of the key range, because then it will be one point, which is not possible with the open half-intervals

Denchick (approved):

PR #277, opened 2023-08-15
ping router before adding to qdb

diPhantxm:


coordinator/provider/coordinator.go
```diff
@@ -915,6 +915,18 @@ func (qc *qdbCoordinator) RegisterRouter(ctx context.Context, r *topology.Router
...
+
+	// ping router
+	conn, err := DialRouter(r)
+	if err != nil {
+		return err
+	}
+	cl := routerproto.NewTopologyServiceClient(conn)
+	_, err = cl.GetRouterStatus(ctx, &routerproto.GetRouterStatusRequest{})
+	if err != nil {
+		return err
```
Denchick, line 927: Maybe wrap this error? Something like `failed to ping router: err`

Denchick, line 927: Maybe wrap this error? Something like failed to ping router: err

Denchick (changes requested):

Denchick (approved):

PR #282, opened 2023-08-21
fix coordinator crash when moving key range twice

diPhantxm:
Why do we mark status of MoveKeyRange as completed in qdb instead of just removing it? Completed operations just pollute qdb, don't they?


reshke:
We do not remove completed operations to track key range movements and use this statistic to decide key-range configuration more efficiently and precise. This is a big TODO and need to be discussed and carefully implemented in future.

reshke (approved):

PR #283, opened 2023-08-21
add checks to move key range

diPhantxm:
* Check if key range exists
* Add tests for non-existent key range scenario and non-existent shard scenario
* Fixed typo

If we encounter an error when moving data from shard to shard, record in qdb will be marked as completed anyway. As stated in #282, we need these records to "decide key-range configuration more efficiently and precise", but actually there was no data movement. It may lead us to wrong decision. What you think?

### DIFF
--- a/test/feature/features/move.feature
+++ b/test/feature/features/move.feature
@@ -164,7 +164,7 @@ Feature: Move test
   Scenario: Move to non-existent shard fails
     When I run SQL on host "coordinator"
     """
-    MOVE KEY RANGE krid1 TO sh3
+    MOVE KEY RANGE krid1 TO non-existent
     """
     Then command return code should be "1"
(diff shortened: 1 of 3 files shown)


test/feature/features/move.feature
```diff
@@ -172,4 +159,26 @@ Feature: Move test
...
     """
     002
-    """
\ No newline at end of file
+    """
+
+  Scenario: Move to non-existent shard fails
+    When I run SQL on host "coordinator"
+    """
+    MOVE KEY RANGE krid1 TO sh3
```
Denchick, line 167: Well, there is a high probability that sh3 will be present in future tests 😄
Let's name it something like `non-existent shard`

Denchick (changes requested):

Denchick (approved):

PR #285, opened 2023-08-23
Parallelize scenarios

nikifkon:
Currently feature tests run in parallel in CI. But this take about 20 min, because a big feature file (e.g. Coordinator.feature). This pull request split big feature files automatically at runtime.

I have a question:
1) Is it OK to use external dependency https://github.com/bbc/cucumber-slicer.git and patch it with patch file?

Also there are a problem I discovered:
1) Overhead of creating new jobs is about 4 minutes, so if we will have a lot feature files in the future, it would be better to not split them.
2) This repo seems only run 20 jobs in parallel, while in my fork it I can run 40 jobs

### DIFF
--- a/.github/scenario_balancer.py
+++ b/.github/scenario_balancer.py
@@ -1,0 +1,17 @@
+N = 16
+import sys
+import os
+
+l = [[int(line.split()[0]), line.split()[1]] for line in sys.stdin.readlines()]
+buckets = [[0, []] for _ in range(N)]
+for w, p in l:
+    b = min(buckets, key=lambda b: b[0])
+    b[0] += w
+    b[1].append(p)
+matrix = []
+for _, ps in buckets:
+    if len(ps) == 0:
+        continue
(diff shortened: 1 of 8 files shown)


nikifkon:
> Overhead of creating new jobs is about 4 minutes, so if we will have a lot feature files in the future, it would be better to not split them.

I wrote [simple script](https://github.com/pg-sharding/spqr/pull/285/files#diff-e1ce4fc02d5ec46e964487b0bc781abd604ac89fb003414fe34889b01c8f623cR1) to balance scenarios onto available jobs (16). So there is no overhead now. Each job utiized according to plan created by the script.

Also I cached `spqr-shard-image`, so it will be reused in all pull request if `docker/shard` folder doesn't change

reshke:
> Is it OK to use external dependency https://github.com/bbc/cucumber-slicer.git and patch it with patch file?

IMO no problem

reshke:
LGTM 💯 💯 💯

reshke (approved):

Issue #279, opened 2023-08-17
Run coordinator as routers part


reshke:
In the begging of SPQR project, the decision to have coordinator as separate unit was made. 

This was not a right, but neither a wrong decision. On one hand, this helps to manage installation topology and keep routers metadata in sync, avoid data loss and routing schema corruption. 

On other hand, why should we have this unit as separate binary being run by admin, robot, etc? 

Maybe run this as part of router process will be better and will simplify SPQR a lot for users. 

TLDR:
simply run  this [function](https://github.com/pg-sharding/spqr/blob/master/cmd/coordinator/main.go#L22) from [here](https://github.com/pg-sharding/spqr/blob/master/cmd/router/main.go#L169) in separate goroutine

Denchick:
I like the idea. We discussed with @x4m, @munakoiso, and @EinKrebs how we should roll out the coordinator. We concluded that it would be easier to combine the coordinator and the router. We haven't decided where the QDB will be launched yet, but by combining these pieces, we can reap the benefits.

Anyway, there are some disadvantages in this desicion, right?
The only thing that confuses me is that @diphantxm has already invested time into fixing bugs and writing tests specifically for the coordinator. If we merge the coordinator into the router now, it might seem like some of that work has gone to waste.

reshke:
>it might seem like some of that work has gone to waste.

Dont think any work is going to waste at all.

PR #287, opened 2023-08-24
Run coordinator together with router

diPhantxm:
#279


pkg/config/data_transfers.go
```diff
@@ -17,11 +17,10 @@ type DatatransferConnections struct {
 }
 
 type ShardConnect struct {
-	Host     string `json:"host" toml:"host" yaml:"host"`
-	Port     string `json:"port" toml:"port" yaml:"port"`
```
reshke, line 21: why `port` here is removed?

test/feature/features/multicoordinator.feature
```diff
@@ -18,38 +18,23 @@ Feature: Coordinator test
...
     """
     SHOW routers
     """
     Then command return code should be "0"
     And SQL result should match regexp
     """
     router
-    """
-
-    Scenario: first coordinator awaits after recovery
```
Denchick, line 39: Why?

pkg/config/data_transfers.go
Denchick: >  that's why I changed host from string to []string and removed port

Nice but `Hosts` then

test/feature/features/coordinator.feature
```diff
@@ -441,37 +447,6 @@ Feature: Coordinator test
     context deadline exceeded
     """
 
-  Scenario: Coordinator can restart
```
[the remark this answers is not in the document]
diPhantxm: At least for now we cannot restart just coordinator, because it's now a part of a router, and not a separate service. There is a scenario for restarting router down below

test/feature/features/multicoordinator.feature
diPhantxm: Because host with coordinator might change during test. After cluster is up or any router is down, new coordinator connection is established. Because maybe it was a router with running coordinator. That's why we cannot check this anymore.
This test assumes we have coordinator on router-1 and coordinator on router-2, and it assumes it remains true after restart of any coordinator. But this is not always true

reshke:
We do not need coordinator console, since we only interact with coordinator using grpc-API in code, and i think that it is acceptable that final users can access spqr installation for different type of requests though router console. 

We need to run admin console in proxy way for several type of queries: 

1) All modify requests to router console should be proxied to coordinator via particular gRPC-call. It is very dangerous to allow user to modify router metadata, like key-ranges or sharding rules, etc. This only should by done by coordinator or via coord. API.

2) All `show` requests, like `SHOW CLIENTS` etc should be processed locally. In future, we can support installation-wide 
show requests processing using additional syntax, like `SHOW CLIENTS` is processed by router but `SHOW GLOBAL CLIENTS` to be proccesed by coordinator (coord. can process such queries already).

3) Do we need to drop coordinator binary? I'm not sure whether it's necessary, since only one coordinator can run simultaneously. In addition, running coordinator from binary may be helpful for debug.

diPhantxm:
> We do not need coordinator console, since we only interact with coordinator using grpc-API in code, and i think that it is acceptable that final users can access spqr installation for different type of requests though router console.
> 
> We need to run admin console in proxy way for several type of queries:
> 
> 1. All modify requests to router console should be proxied to coordinator via particular gRPC-call. It is very dangerous to allow user to modify router metadata, like key-ranges or sharding rules, etc. This only should by done by coordinator or via coord. API.
> 2. All `show` requests, like `SHOW CLIENTS` etc should be processed locally. In future, we can support installation-wide
>    show requests processing using additional syntax, like `SHOW CLIENTS` is processed by router but `SHOW GLOBAL CLIENTS` to be proccesed by coordinator (coord. can process such queries already).
> 3. Do we need to drop coordinator binary? I'm not sure whether it's necessary, since only one coordinator can run simultaneously. In addition, running coordinator from binary may be helpful for debug.

Sounds great. I will do it ASAP

pkg/config/router.go
```diff
@@ -60,6 +62,8 @@ type Router struct {
 	ShardMapping     map[string]*Shard `json:"shards" toml:"shards" yaml:"shards"`
 
 	ReusePort bool `json:"reuse_port" toml:"reuse_port" yaml:"reuse_port"`
+
+	WithCoordinator bool `json:"with_coordinator" toml:"with_coordinator" yaml:"with_coordinator"`
```
reshke, line 66: I would prefer `DispatchCoordinator`

reshke (approved):
Overall its ok
lgtm

reshke:
Have enhanced commit message(squashed) with detailed explaining of this PR.

docs/Router.md
```diff
@@ -2,7 +2,7 @@
 
 ![Router schema](router.jpg "Router")
 
-Consist of 3 parts:
+Consist of 4 parts:
```
Denchick, line 5: typo

Denchick (approved):

PR #294, opened 2023-09-05
Logproxy

debebantur:
USAGE EXAMPLE:
Precondition:
you have dbs "postgres" and "db1" running on localhost:5432
Steps:
1) Start proxy 
./logproxy run -H localhost -p 5432 -l testLogs1.txt
2) send requests using
psql -h localhost  -p 5433 -U postgres  -d postgres
3) replay requests
./logproxy replay -U postgres -d db1 -l testLogs.txt


test/logproxy/logproxy.go
```diff
@@ -47,79 +56,330 @@ func (p *Proxy) Run() error {
 
 	go accept(listener, cChan)
 
+	spqrlog.Zero.Info().Msg("Proxy is up and listening on port 5433")
```
reshke, line 59: port should be configurable

test/logproxy/logproxy.go
```diff
@@ -47,79 +56,330 @@ func (p *Proxy) Run() error {
...
+	fm := &pgproto3.Query{}
+	err = fm.Decode(msg)
+	if err != nil {
+		return time.Now(), nil, err
+	}
+
+	return ti, fm, nil
+}
+
+func startup(netconn net.Conn, frontend *pgproto3.Frontend, cl *pgproto3.Backend) error {
```
reshke, line 275: there is same exact function in router/client
we should reuse code, not write it twice

PR #327, opened 2023-11-27
BEGIN tests

CBists:


Denchick:
Add please more queries for BEGIN
https://www.postgresql.org/docs/current/sql-begin.html

Denchick (approved):

PR #328, opened 2023-11-28
Support more spqr specific params, support scatter route hint

reshke:


router/client/spqr_param.go
```diff
@@ -0,0 +1,7 @@
+package client
+
+const (
+	SPQR_DEFAULT_ROUTE_BEHAVIOUR = "__spqr_default_route_behaviour"
```
CBists, line 4: may be use `__spqr__` prefix to our variable, like in SET command?

EinKrebs (approved):

PR #344, opened 2023-12-05
Fix unittests and etcd paths

Denchick:


router/server/server.go
```diff
@@ -24,6 +24,7 @@ type Server interface {
 	AddDataShard(clid string, shardKey kr.ShardKey, tsa string) error
 	UnRouteShard(sh kr.ShardKey, rule *config.FrontendRule) error
 	Datashards() []shard.Shard
+	DatashardIds() []string
```
reshke, line 28: Why do we need another method in server iface? We can write separate utility function which accepts []shard.Shard and returns []string

reshke (changes requested):

router/server/shard.go
```diff
@@ -144,7 +144,7 @@ func (srv *ShardServer) Receive() (pgproto3.BackendMessage, error) {
 		Uint("single-shard", spqrlog.GetPointer(srv)).
 		Type("message-type", msg).
 		Str("txstatus", srv.TxStatus().String()).
-		Str("shard", srv.Datashards()[0].ID()).
+		Strs("shards", srv.DatashardIds()).
```
EinKrebs, line 147:
```suggestion
		Strs("shards", srv.DatashardIds()[0]).
```
Зачем убирать индекс?

router/relay/relay.go
```diff
@@ -608,7 +608,7 @@ func (rst *RelayStateImpl) ProcQuery(query pgproto3.FrontendMessage, waitForResp
 	}
 
 	spqrlog.Zero.Debug().
-		Str("server", server.Datashards()[0].ID()).
+		Strs("shards", server.DatashardIds()).
```
EinKrebs, line 611:
```suggestion
		Strs("shards", server.DatashardIds()[0]).
```

router/relay/relay.go
```diff
@@ -918,7 +918,7 @@ func (rst *RelayStateImpl) DeployPrepStmt(qname string) (shard.PreparedStatement
 		Str("query", query).
 		Uint64("hash", hash).
 		Str("client", rst.Client().ID()).
-		Str("shard", rst.Client().Server().Datashards()[0].ID()).
+		Strs("shards", rst.Client().Server().DatashardIds()).
```
EinKrebs, line 921:
```suggestion
		Strs("shards", rst.Client().Server().DatashardIds()[0]).
```

EinKrebs (changes requested):

router/server/shard.go
reshke: thats because its now list of strings, not just one?

router/relay/relay.go
Denchick: panic: runtime error: index out of range

router/server/shard.go
Denchick: panic: runtime error: index out of range

router/server/server.go
Denchick: Add util function

reshke (approved):

PR #355, opened 2023-12-11
fix feature

CBists:


pkg/coord/adapter.go
```diff
@@ -65,8 +64,28 @@ func (a *adapter) ListAllKeyRanges(ctx context.Context) ([]*kr.KeyRange, error)
 }
 
 func (a *adapter) AddKeyRange(ctx context.Context, kr *kr.KeyRange) error {
+	_, err := a.GetShardInfo(ctx, kr.ShardID)
+	if err != nil {
+		return err
+	}
+
+	existDataspace, err := a.ListDataspace(ctx)
```
EinKrebs, line 72: Consider renaming existDataspace to dataspaces

pkg/meta/meta.go
```diff
@@ -174,8 +173,7 @@ func processCreate(ctx context.Context, astmt spqrparser.Statement, mngr EntityM
 		return cli.AddShardingRule(ctx, shardingRule)
 	case *spqrparser.KeyRangeDefinition:
 		req := kr.KeyRangeFromSQL(stmt)
-		if err := ops.AddKeyRangeWithChecks(ctx, mngr.QDB(), req); err != nil {
```
EinKrebs, line 177: Why do so?

EinKrebs (changes requested):

pkg/meta/meta.go
CBists: Because AddKeyRangeWithChecks doesn't work with adapter and after we are calling this method router shutting down with error.

EinKrebs (approved):

pkg/coord/adapter.go
```diff
@@ -65,8 +64,28 @@ func (a *adapter) ListAllKeyRanges(ctx context.Context) ([]*kr.KeyRange, error)
...
 func (a *adapter) AddKeyRange(ctx context.Context, kr *kr.KeyRange) error {
+	_, err := a.GetShardInfo(ctx, kr.ShardID)
+	if err != nil {
+		return err
+	}
+
+	dataspaces, err := a.ListDataspace(ctx)
+	if err != nil {
+		return err
+	}
```
reshke, line 75: why do we execute this checks here if we have them in `AddKeyRangeWithChecks`?

reshke (changes requested):

pkg/coord/adapter.go
CBists: Because we don't have QDB struct in adapter

PR #361, opened 2023-12-12
Pre-defined hash functions for key range

reshke:


router/qrouter/proxy_routing.go
```diff
@@ -158,15 +159,23 @@ func (qr *ProxyQrouter) DeparseKeyWithRangesInternal(ctx context.Context, key st
...
+			return nil, err
+		}
+		return qr.DeparseKeyWithRangesInternal(ctx, string(hashedKey), meta)
 	case *lyx.AExprConst:
-		return qr.DeparseKeyWithRangesInternal(ctx, e.Value, meta)
+		hashedKey, err := hashfunction.ApplyHashFunction([]byte(e.Value), hf)
+		if err != nil {
+			return nil, err
+		}
+		return qr.DeparseKeyWithRangesInternal(ctx, string(hashedKey), meta)
```
EinKrebs, line 178: You can move key hashing before the switch

router/qrouter/proxy_routing.go
```diff
@@ -647,40 +669,49 @@ func (qr *ProxyQrouter) routeWithRules(ctx context.Context, stmt lyx.Node, datas
 			meta.offsets = offsets
 			routed := false
 			if len(meta.offsets) != 0 && len(meta.TargetList) > meta.offsets[0] {
-
-				currroute, err := qr.RouteKeyWithRanges(ctx, meta.TargetList[meta.offsets[0]], meta)
+				hf, err := hashfunction.HashFunctionByName(rule.Entries[0].HashFunction)
```
EinKrebs, line 672: Duplicate code here too

EinKrebs (changes requested):

router/qrouter/proxy_routing.go
reshke: here we cannot - look closer, meta.params[e.Number] vs e.Value

router/qrouter/proxy_routing.go
reshke: line 672 duplicated with which one?

EinKrebs: With line 705

EinKrebs (approved):

Issue #319, opened 2023-11-08
Change error message when happend routing error


CBists:
When router can not route request, a parsing error message is sent to the interface
ERROR:  error processing query 'SELECT * FROM xx WHERE w_id = 5;': too complex query to parse

PR #374, opened 2023-12-24
fix error from issues

vovotata:


vovotata:
@Denchick посмотрите пж, мне для курса opensource нужно

CBists (approved):

x4m (approved):
I do not see any problem to change the wording of that error. Though, I think we also should throw some specific error code. This code muse bt known to the client and indicate that query was understood and cannot be dispatched to one shard. Another error code should indicate that query is not understood.
There's no point in relying on messages, they are for postmortem analisys only.

vovotata:
https://github.com/pg-sharding/spqr/issues/319

PR #390, opened 2024-01-11
Add docker container for generate protoc, mock, goyacc and regel

CBists:
To generate packages with the same versions for all users


reshke:
nice

docker-compose.yaml
```diff
@@ -87,3 +87,7 @@ services:
         build:
             dockerfile: ./docker/slicer/Dockerfile
             context: .
+    generator:
+        build:
+            dockerfile: ./docker/generator/Dockerfile
+            context: .
```
EinKrebs, line 93: It's better not to include generator here, for it has no purpose. Better to use Docker directly in Makefile

EinKrebs (changes requested):

docker/generator/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM spqr-base-image
```
EinKrebs, line 1: Maybe use `golang` docker image here, not to install go manually

Makefile
```diff
@@ -159,10 +159,10 @@ yaccgen:
 gen: gogen yaccgen mockgen
 
 generate: build_images
-	docker compose up --remove-orphans generator
+	docker build -f docker/generator/Dockerfile -t spqr-generator .
+	docker run --name spqr-generator-1 spqr-generator --rm
```
EinKrebs, line 163: You need to remove `--rm` flag

EinKrebs (changes requested):

Makefile
EinKrebs: Also add `docker container rm spqr-generator-1` to the end of `generate` target

EinKrebs (approved):

PR #416, opened 2024-01-18
Drop sharding rule, add relation definitions to dataspace

reshke:
In this pr we replace current routing logic based on sharding rules with new one, based on dataspaces. 

refer
https://github.com/pg-sharding/spqr/discussions/431


EinKrebs:
Consider deprecating sharding rules instead of dropping them

reshke:
sharding rules dropped from SQL api, however they still present in grpc API and in backend code

EinKrebs (approved):

router/qrouter/proxy_routing.go
```diff
@@ -192,7 +189,8 @@ func (qr *ProxyQrouter) RouteKeyWithRanges(ctx context.Context, expr lyx.Node, m
 		spqrlog.Zero.Debug().Str("key", string(meta.params[e.Number-1])).Str("hashed key", string(hashedKey)).Msg("applying hash function on key")
 
 		return qr.DeparseKeyWithRangesInternal(ctx, string(hashedKey), meta)
-	case *lyx.AExprConst:
+	case *lyx.AExprIConst:
+		if
```
EinKrebs, line 193: ??

EinKrebs (changes requested):

router/qrouter/proxy_routing.go
EinKrebs: Why if here?

EinKrebs, line 193: Incorrect if

EinKrebs (changes requested):

router/qrouter/proxy_routing.go
reshke: yea...
just pushed a bunch of work

reshke:
I would like to propose to use only one bound in key range (lower one). To check if given key lays within given keyrange we will only have to check that given key is greater (>=) then KR bound and no other (greater) key range satisfies the same.

EinKrebs:
@reshke PR is obsolete, consider closing it

PR #427, opened 2024-01-19
Fix xproto tests

EinKrebs:


Denchick (changes requested):
FAIL	github.com/pg-sharding/spqr/test/xproto	0.031s

Denchick (approved):

PR #454, opened 2024-01-25
Coordinator auth

Denchick:
Seem working

```
➜  spqr git:(make-coordinator-auth-again) ✗ psql "host=localhost port=7002 dbname=spqr-console user=spqr-console"
Password for user spqr-console: 
psql (14.10 (Homebrew), server 0.0.0)
Type "help" for help.

spqr-console=> SHOW sharding_rules;
 Sharding Rule ID | Dataspace ID | Table Name | Columns | Hash Function 
------------------+--------------+------------+---------+---------------
(0 rows)

spqr-console=> ^C
spqr-console=> 
\q
➜  spqr git:(make-coordinator-auth-again) ✗ psql "host=localhost port=7002 dbname=spqr-console user=spqr-consoled"
^C
```

### DIFF
--- a/coordinator/provider/coordinator.go
+++ b/coordinator/provider/coordinator.go
@@ -1089,20 +1089,14 @@ func (qc *qdbCoordinator) PrepareClient(nconn net.Conn) (CoordinatorClient, erro
 		return cl, nil
 	}
 
-	coordCfg := config.CoordinatorConfig()
-
-	if cl.DB() != coordCfg.Dbname || cl.Usr() != coordCfg.Username {
-		return cl, fmt.Errorf("try to connect to '%s' db, skipping", cl.DB())
-	}
-
 	spqrlog.Zero.Info().
 		Str("user", cl.Usr()).
 		Str("db", cl.DB()).
 		Msg("initialized client connection")
 
 	var authRule *config.AuthCfg
-	if coordCfg.Auth != nil {
-		authRule = coordCfg.Auth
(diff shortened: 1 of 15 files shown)


coordinator/provider/coordinator.go
```diff
@@ -1078,23 +1081,37 @@ func (qc *qdbCoordinator) UnregisterRouter(ctx context.Context, rID string) erro
...
 		return nil, err
 	}
 
 	if cl.CancelMsg() != nil {
 		return cl, nil
 	}
 
+	coordCfg := config.CoordinatorConfig()
+
+	if cl.DB() != coordCfg.Dbname || cl.Usr() != coordCfg.Username {
```
reshke, line 1094: why exaclty?

reshke: maybe enforce db/usr params for route instread?

test/regress/conf/coordinator.yaml
```diff
@@ -1,5 +1,7 @@
+log_level: info
 host: regress_coordinator
+dbname: spqr-console
```
reshke, line 3: lets remove this and enforce in code instead. 
there no reason for user to think about usr/db when it connects to coordinator. 
It makes a lot of sense in router case, since this params is actually used for routing there.

reshke (changes requested):
+-OK

reshke (approved):

PR #477, opened 2024-02-06
Removed deprecated grpc.WithInsecure method from etcdqdb.go

EinKrebs:


qdb/etcdqdb.go
```diff
@@ -30,8 +31,8 @@ var _ QDB = &EtcdQDB{}
 func NewEtcdQDB(addr string) (*EtcdQDB, error) {
 	cli, err := clientv3.New(clientv3.Config{
 		Endpoints: []string{addr},
-		DialOptions: []grpc.DialOption{ // TODO remove WithInsecure
-			grpc.WithInsecure(), //nolint:all
+		DialOptions: []grpc.DialOption{
+			grpc.WithTransportCredentials(insecure.NewCredentials()),
```
Denchick, line 35: Do you have proof that it is working? Because there are no grpc tests

PR #480, opened 2024-02-06
Added 'ALTER DISTRIBUTION ATTACH RELATION' command

EinKrebs:


yacc/console/gram.y
```diff
@@ -120,8 +124,8 @@ func randomHex(n int) (string, error) {
 // routers
 %token <str> SHUTDOWN LISTEN REGISTER UNREGISTER ROUTER ROUTE
 
-%token <str> CREATE ADD DROP LOCK UNLOCK SPLIT MOVE COMPOSE SET CASCADE ATTACH
-%token <str> SHARDING COLUMN TABLE HASH FUNCTION KEY RANGE DISTRIBUTION
+%token <str> CREATE ADD DROP LOCK UNLOCK SPLIT MOVE COMPOSE SET CASCADE ATTACH ALTER
```
reshke, line 127: maybe drop attach keyword and grammar?

EinKrebs: I don't think so, don't we need to implement relation detaching in future?

EinKrebs: I don't think so, don't we need to implement relation detaching in future?

Denchick (approved):

PR #484, opened 2024-02-06
Added 'ALTER DISTRIBUTION ds_id DETACH RELATION rel_id' statement support

EinKrebs:


protos/distribution.proto
```diff
@@ -41,6 +41,13 @@ message AlterDistributionAttachRequest{
 
 message AlterDistributionAttachReply{}
 
+message AlterDistributionDetachRequest{
+  string id = 1;
+  string relName = 2;
```
reshke, line 46: Maybe repeated string?

reshke (approved):

PR #490, opened 2024-02-12
Forbid default distribution & coordinator fixes

EinKrebs:


Makefile
```diff
@@ -89,7 +89,7 @@ regress_local: proxy_2sh_run
 	./script/regress_local.sh
 
 regress: build_images
-	docker compose -f test/regress/docker-compose.yaml up --remove-orphans --force-recreate --exit-code-from regress --build coordinator router shard1 shard2 regress qdb01
+	docker compose -f test/regress/docker-compose.yaml down && docker compose -f test/regress/docker-compose.yaml run --build --remove-orphans regress
```
reshke, line 92: I'm not sure running test this way is idempotent.
Will all data inserted in first test run affect next one?

EinKrebs: No, it won't, as `docker compose down` actually removes containers

reshke (approved):

PR #508, opened 2024-02-16
Enhace yacc parser to accept multiple relations

reshke:


yacc/console/gram.y
```diff
@@ -493,16 +497,28 @@ distribution_key_entry:
...
+			Name: 	 $2,
+			DistributionKey: $5,
 		}
 	}
 
 
+distributed_relation_list_def:
+	distributed_relation_def {
+		$$ = []*DistributedRelation{$1}
+	} | distributed_relation_list_def distributed_relation_def {
```
EinKrebs, line 513:
```suggestion
	} | distributed_relation_list_def TCOMMA distributed_relation_def {
```
Might use brackets as well

EinKrebs (changes requested):

yacc/console/gram.y
reshke: we already have comma in DISTRIBUTION KEY clause. Relations are separated with RELATION token. 

Like this
```

		ALTER DISTRIBUTION 
			ds1 
		ATTACH
			RELATION t DISTRIBUTION KEY id1, id2 HASH FUNCTION murmur
			RELATION t2 DISTRIBUTION KEY xd1, xd2 HASH FUNCTION city
```

EinKrebs (approved):

Issue #512, opened 2024-02-21
Relations List via SQL


Denchick:
Add view to see list of relation in dataspace to administrator console. Something like: `SHOW relations [WHERE distirbution = d_id]`

PR #514, opened 2024-02-21
More verbose output in 'SHOW distributions' command

EinKrebs:


reshke:
Also all distributed relations?

EinKrebs:
> Also all distributed relations?

Planned as separate command in #512

reshke (approved):

PR #557, opened 2024-03-17
added tests to shard utils ShardIDs

IakimovNutria:


pkg/shard/mock_shard.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"crypto/tls"
+	"github.com/jackc/pgx/v5/pgproto3"
+	"github.com/pg-sharding/spqr/pkg/config"
+	"github.com/pg-sharding/spqr/pkg/conn"
+	"github.com/pg-sharding/spqr/pkg/models/kr"
+	"github.com/pg-sharding/spqr/pkg/txstatus"
+)
+
+type mockShard struct {
```
EinKrebs, line 12: please use `MockShard` from `mock` package

EinKrebs (changes requested):

pkg/shard/mock_shard.go
IakimovNutria: I can't because it causes import cycle

reshke: anyway you should use mockgen.
What is the import cycle are you struggling with?

IakimovNutria: > What is the import cycle are you struggling with?

package github.com/pg-sharding/spqr/pkg/shard
	imports github.com/pg-sharding/spqr/pkg/mock/shard
	imports github.com/pg-sharding/spqr/pkg/shard: import cycle not allowed in test

EinKrebs: Consider changing package name in your test to shard_test

EinKrebs (approved):

PR #562, opened 2024-03-19
Add flag to configure memqdb persistent setup

reshke:


cmd/router/main.go
```diff
@@ -66,6 +68,7 @@ func init() {
 
 	rootCmd.PersistentFlags().StringVarP(&ccfgPath, "coordinator-config", "", "/etc/spqr/coordinator.yaml", "path to coordinator config file")
 	rootCmd.PersistentFlags().StringVarP(&qdbImpl, "qdb-impl", "", "etcd", "which implementation of QDB to use.")
+	rootCmd.PersistentFlags().BoolVarP(&persist, "persist", "", false, "tells router to persistits configuration in non-clustered setup")
```
EinKrebs, line 71:
```suggestion
	rootCmd.PersistentFlags().BoolVarP(&persist, "persist", "", false, "tells router to persist its configuration in non-clustered setup")
```

EinKrebs (changes requested):

EinKrebs (approved):

PR #568, opened 2024-03-24
Set sslmode: allow in regress test

NitroLine:
Add localhost certs

Set it in config for router

Set PGSSLMODE to allow and ssmode: require in router regress tests


docker/router/ssl/localhost.crt
```diff
@@ -0,0 +1,19 @@
+-----BEGIN CERTIFICATE-----
```
munakoiso, line 1: The certificate will expire on April 22 and these tests will begin to fail.
let's change it to something that expires somewhere in 1000 years

```
> openssl x509 -text -noout -in foo
Certificate:
...
Validity
...
            Not After : Apr 22 18:58:31 2024 GMT
```

munakoiso (changes requested):

docker/router/ssl/localhost.crt
NitroLine: Now its will expire in Jul 27 09:18:42 3023 GMT

docker/router/cfg.yaml
```diff
@@ -10,7 +10,7 @@ time_quantiles:
 frontend_tls:
   key_file: /etc/odyssey/ssl/server.key
   cert_file: /etc/odyssey/ssl/server.crt
```
EinKrebs, line 12:
```suggestion
  key_file: /etc/spqr/ssl/server.key
  cert_file: /etc/spqr/ssl/server.crt
```

EinKrebs (changes requested):

EinKrebs (approved):

munakoiso (approved):

PR #573, opened 2024-03-27
More thorough check for relation detachment when distribution dropped

EinKrebs:


reshke (approved):

Denchick (approved):

test/feature/spqr_test.go
```diff
@@ -725,6 +726,21 @@ func (tctx *testContext) stepRecordQDBTx(key string, body *godog.DocString) erro
 	return tctx.qdb.RecordTransferTx(context.TODO(), key, &st)
 }
 
+func (tctx *testContext) stepQDBShouldNotContainRelation(key string) error {
+	_, err := tctx.qdb.GetRelationDistribution(context.TODO(), key)
+	switch t := err.(type) {
+	case *spqrerror.SpqrError:
+		if t.ErrorCode == spqrerror.SPQR_NO_DISTRIBUTION {
```
Denchick, line 733: But if there is no distribution, qdb still may contain relation, doesn't it?

Denchick, line 733: But if there is no distribution, qdb still may contain relation, doesn't it?

Denchick (changes requested):

test/feature/spqr_test.go
EinKrebs: Attempt to address such table will result in an error anyway, plus we prevent deletion of distributions with attached realtions, so I don't think it does

Denchick (approved):

PR #590, opened 2024-04-02
Init from etcd: basic support

reshke:
change router initialization in clustered setup mechanics.
Router requests current metadata setup from qdb on start. Router sync metadata routine now sync only current coordinator address for router proxy queries.


Denchick (changes requested):
It would be nice to add tests that check whether or not the initialization of metadata for different types of installations actually works.

examples/coordinator-init.yaml
```diff
@@ -0,0 +1,66 @@
+log_level: debug
```
Denchick, line 1: it looks like a router config, is it definitely coordinator-init.yaml?

Denchick (approved):

Denchick (changes requested):
Add comments

cmd/router/main.go
```diff
@@ -317,6 +317,26 @@ var runCmd = &cobra.Command{
 			}
 		}()
 
+		/* initialize metadata */
```
Denchick, line 320:
```suggestion
		// The router will execute the file "init_sql" before starting.
```

Denchick (approved):

examples/coordinator-init.yaml
reshke: it is router-etcd-init.yaml

cmd/router/main.go
reshke: no, this is both init sql and coodinator init cases here

PR #612, opened 2024-04-12
fix router.yaml config file

JoBeR007:


examples/router.yaml
```diff
@@ -52,21 +58,11 @@ shards:
     pwd: 12345678
     type: DATA
     hosts:
-      - '192.168.233.2:6432'
+      - 'localhost:5550'
   sh2:
     tls:
       root_cert_file: /path/to/root.ca
       sslmode: disable
     type: DATA
```
EinKrebs, line 66: Please add db and usr here

EinKrebs (changes requested):

examples/router.yaml
JoBeR007: Added in the last commit

EinKrebs (approved):

PR #625, opened 2024-04-17
Process several set values for each distribution key

EinKrebs:


reshke (approved):

router/qrouter/proxy_routing_test.go
```diff
@@ -1329,3 +1395,7 @@ func TestMiscRouting(t *testing.T) {
 		}
 	}
 }
+
+func TestSimple(t *testing.T) {
```
reshke, line 1399: ?

reshke (changes requested):

reshke (approved):

PR #630, opened 2024-04-22
Fix double take of acquired connections in shardpool.go

EinKrebs:


pkg/pool/shardpool.go
```diff
@@ -169,9 +171,10 @@ func (h *shardPool) Connection(
 		return nil, err
 	}
 
-	h.mu.Lock()
-	defer h.mu.Unlock()
-
+	if _, ok := h.active[sh.ID()]; ok {
+		// double take
+		panic(sh)
```
Denchick, line 176: Its always better not to panic

PR #642, opened 2024-04-27
LDAP auth for frontend router

NitroLine:


test/feature/docker-compose.yaml
```diff
@@ -122,6 +123,21 @@ services:
     ports:
       - "2379:2379"
 
+  openldapr:
```
Denchick, line 126: Why not just `openldap`?

Denchick (approved):
Thanks for your contribution!

PR #646, opened 2024-05-02
Added test of multiparam prepared statement

JoBeR007:


test/xproto/proto_test.go
```diff
@@ -102,10 +102,6 @@ func SetupSharding() {
 	if err != nil {
 		_, _ = fmt.Fprintf(os.Stderr, "could not setup sharding: %s\n", err)
 	}
-	_, err = conn.Exec(context.Background(), "CREATE SHARDING RULE r1 COLUMNS id FOR DISTRIBUTION ds1;")
```
reshke, line 105: can be a separate change

EinKrebs, line 105: I agree, please consider moving it to separate PR

EinKrebs (changes requested):

PR #664, opened 2024-05-20
 Able to use different username when internal backend connect to shards

NitroLine:
-Add a different auth configuration struct for backend auth rules (because AuthMethod is not used here).
-Add the ability to use different usernames when internal backend connects to shards. 
This may be used when:
- Different shards have different PostgreSQL usernames.
- When using LDAP/GSS auth - SPQR as one entrypoint for many LDAP users able to connect.

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -44,7 +44,7 @@ func AuthBackend(shard conn.DBInstance, berule *config.BackendRule, msg pgproto3
 		return nil
 	case *pgproto3.AuthenticationMD5Password:
 
-		var rule *config.AuthCfg
+		var rule *config.AuthBackendCfg
 		if berule.AuthRules == nil {
 			rule = berule.DefaultAuthRule
 		} else if _, exists := berule.AuthRules[shard.ShardName()]; exists {
@@ -56,7 +56,10 @@ func AuthBackend(shard conn.DBInstance, berule *config.BackendRule, msg pgproto3
 		if rule == nil {
 			return fmt.Errorf("auth rule not set for %s-%s-%s", shard.ShardName(), berule.DB, berule.Usr)
 		}
-
+		username := berule.Usr
+		if rule != nil && rule.Usr != "" {
+			username = rule.Usr
+		}
 		var res []byte
 
 		/* password may be configured in partially-calculated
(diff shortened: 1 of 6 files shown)


pkg/auth/auth.go
```diff
@@ -56,7 +56,10 @@ func AuthBackend(shard conn.DBInstance, berule *config.BackendRule, msg pgproto3
 		if rule == nil {
 			return fmt.Errorf("auth rule not set for %s-%s-%s", shard.ShardName(), berule.DB, berule.Usr)
 		}
-
+		username := berule.Usr
+		if rule != nil && rule.Usr != "" {
```
EinKrebs, line 60:
```suggestion
		if rule.Usr != "" {
```
`rule != nil` is redundant

pkg/auth/auth.go
```diff
@@ -101,15 +104,22 @@ func AuthBackend(shard conn.DBInstance, berule *config.BackendRule, msg pgproto3
...
 			rule = berule.AuthRules[shard.ShardName()]
 		} else {
 			rule = berule.DefaultAuthRule
 		}
-		clientSHA256, err := scram.SHA256.NewClient(berule.Usr, rule.Password, "")
+		if rule == nil {
+			return fmt.Errorf("auth rule not set for %s-%s-%s", shard.ShardName(), berule.DB, berule.Usr)
+		}
+		username := berule.Usr
+		if rule != nil && rule.Usr != "" {
```
EinKrebs, line 119:
```suggestion
		if rule.Usr != "" {
```
Same here

pkg/auth/auth_test.go
```diff
@@ -243,8 +243,8 @@ func TestErrorWhenNoPasswordForShard(t *testing.T) {
 // - None.
 func TestCanConnectWithDefaultRule(t *testing.T) {
 	//init test data
-	authRule_md5 := &config.AuthCfg{
-		Method:   "md5",
+	authRule_md5 := &config.AuthBackendCfg{
+		// Method:   "md5",
```
EinKrebs, line 247: Don't comment this, delete it instead

EinKrebs (changes requested):

EinKrebs (approved):

PR #669, opened 2024-05-25
LDAP auth

lGreenLightl:
### LDAP auth

Other implementation of LDAP auth with new features:

- _2 binding modes_
- _Multiple servers support_
- _TLS support (one-way and two-way)_

Also changed LDAP server docker image (currently supported) for feature tests


Denchick (approved):

pkg/auth/auth.go
```diff
@@ -321,79 +318,59 @@ func AuthFrontend(cl client.Client, rule *config.FrontendRule) error {
 		err = cl.Send(&pgproto3.AuthenticationSASLFinal{Data: []byte(finalMsg)})
 		return err
 	case config.AuthLDAP:
-		if rule.AuthRule.LDAPConfig == nil {
```
EinKrebs, line 324: Why did you remove this check?

EinKrebs (changes requested):

pkg/auth/auth.go
lGreenLightl: I forgot to add it back when I removed previous code. Fixed

EinKrebs (approved):

PR #706, opened 2024-07-09
Make FOR DISTRIBUTION optional when there is only one distribution

visill:


qdb/ops/ops.go
```diff
@@ -19,6 +19,20 @@ func CreateKeyRangeWithChecks(ctx context.Context, qdb qdb.QDB, keyRange *kr.Key
...
+	if keyRange.Distribution == "default" {
+		list, err := qdb.ListDistributions(ctx)
+		if err != nil {
+			return spqrerror.New(spqrerror.SPQR_NO_DISTRIBUTION, "error while selecting list of distributions")
+		}
+		if len(list) == 0 {
+			return spqrerror.New(spqrerror.SPQR_NO_DISTRIBUTION, "you didn't have any distributions")
+		}
+		if len(list) != 1 {
+			return spqrerror.New(spqrerror.SPQR_NO_DISTRIBUTION, "distributions count not equal one, use DISTRIBUTION BY syntax")
```
EinKrebs, line 31:
```suggestion
			return spqrerror.New(spqrerror.SPQR_NO_DISTRIBUTION, "distributions count not equal one, use FOR DISTRIBUTION syntax")
```

test/regress/tests/console/expected/add.out
```diff
@@ -5,19 +5,23 @@
...
 
+CREATE KEY RANGE krid1 FROM 1 ROUTE TO sh1;
+ERROR:  you didn't have any distributions.
 CREATE DISTRIBUTION ds1 COLUMN TYPES integer;
     add distribution    
 ------------------------
  distribution id -> ds1
 (1 row)
 
-CREATE KEY RANGE krid1 FROM 1 ROUTE TO sh1 FOR DISTRIBUTION ds1;
```
EinKrebs, line 14: Consider adding new tests rather than rewriting existing

EinKrebs (changes requested):

test/regress/tests/console/expected/add.out
reshke: 💯

visill: I changed the logic of creating the key range. We have a test testing the logic of creating key range and distribution. I'm changing it without doing a duplicate test. If you think you need to do a separate test, then I will do a separate key_range test.

PR #720, opened 2024-07-16
Make FOR DISTRIBUTION optional when there is only one distribution

visill:


pkg/meta/meta.go
```diff
@@ -192,6 +195,19 @@ func processCreate(ctx context.Context, astmt spqrparser.Statement, mngr EntityM
 	case *spqrparser.ShardingRuleDefinition:
 		return cli.ReportError(spqrerror.ShardingRulesRemoved)
 	case *spqrparser.KeyRangeDefinition:
+		if stmt.Distribution == "default" {
+			list, err := mngr.ListDistributions(ctx)
+			if err != nil {
+				return spqrerror.New(spqrerror.SPQR_NO_DISTRIBUTION, "error while selecting list of distributions")
+			}
+			if len(list) <= 0 {
+				return spqrerror.New(spqrerror.SPQR_NO_DISTRIBUTION, "you didn't have any distributions")
```
EinKrebs, line 204:
```suggestion
				return spqrerror.New(spqrerror.SPQR_NO_DISTRIBUTION, "you don't have any distributions")
```

pkg/meta/meta.go
```diff
@@ -192,6 +195,19 @@ func processCreate(ctx context.Context, astmt spqrparser.Statement, mngr EntityM
 	case *spqrparser.ShardingRuleDefinition:
 		return cli.ReportError(spqrerror.ShardingRulesRemoved)
 	case *spqrparser.KeyRangeDefinition:
+		if stmt.Distribution == "default" {
+			list, err := mngr.ListDistributions(ctx)
+			if err != nil {
+				return spqrerror.New(spqrerror.SPQR_NO_DISTRIBUTION, "error while selecting list of distributions")
+			}
+			if len(list) <= 0 {
```
EinKrebs, line 203:
```suggestion
			if len(list) == 0 {
```
Can be simplified

test/regress/schedule/console
```diff
@@ -11,3 +11,4 @@ test: hash
 test: teardown
 test: delete_distribution
 test: sharding_rules
+test: key_range
```
EinKrebs, line 14: Consider renaming it to smth like default_distribution

EinKrebs (changes requested):

EinKrebs (approved):

EinKrebs (approved):

PR #744, opened 2024-08-19
Add ORDER BY selecting for clients

visill:


reshke:
general coding done ok.
however test needed here. You can use mock generator to get one

Denchick (changes requested):
What happens if the result is very large (for example, several hundred million rows) and does not fit into the router's memory?

There is 2 options here: 
- do not crash and produce correct result (I have no idea how)
- return an error or explicitly somehow report that the response is too large

And we need tests for extremely large response situation.

visill:
@Denchick 1) https://stackoverflow.com/questions/30577308/golang-cannot-recover-from-out-of-memory-crash we cant

visill:
https://github.com/pg-sharding/spqr/pull/753

PR #770, opened 2024-09-16
Change output type of 'SHOW time_quantiles' to float8

EinKrebs:


reshke (approved):

test/regress/conf/router.yaml
```diff
@@ -3,7 +3,7 @@ router_port: '6432'
 admin_console_port: '7432'
 grpc_api_port: '7000'
 router_mode: PROXY
-log_level: fatal
+log_level: debug
```
Denchick, line 6: The router will work very slow, even tests may flap

Denchick (changes requested):

Denchick (approved):

PR #771, opened 2024-09-16
Group by in Coordinator SHOW

diPhantxm:
Tried to make it in more generic way, just in case we want to group in some other queries.
If group by is specified, column and count is returned


EinKrebs (changes requested):
Check out #773, pls

EinKrebs (approved):

PR #783, opened 2024-10-01
Fix dropping key range in MemQDB

EinKrebs:


reshke (approved):

Denchick (changes requested):
Let's add a unit test for this situation

Denchick (approved):

PR #835, opened 2024-11-22
MulticastUnroutableInsertStatement considered harmfull

reshke:


Denchick (changes requested):
> MulticastUnroutableInsertStatement considered harmfull

Why? I agree that the setting is useless, but I wonder why you think it's harmful?

Denchick (approved):

reshke:
> > MulticastUnroutableInsertStatement considered harmfull
> 
> Why? I agree that the setting is useless, but I wonder why you think it's harmful?

it is indeed harmful for distributed relation. This setting leads to sharding data corruption in case when router dont understand how to handle a query.

PR #855, opened 2024-12-10
replicated distribution --> reference distribution

Denchick:


pkg/conn/instance.go
```diff
@@ -187,7 +187,7 @@ func setTCPUserTimeout(d time.Duration) func(string, string, syscall.RawConn) er
 				#define TCP_USER_TIMEOUT	 18 // How long for loss retry before timeout
 			*/
 
-			sysErr = syscall.SetsockoptInt(int(fd), syscall.SOL_TCP, 0x12,
```
reshke, line 190: wtf

Denchick: some macbook m3 compilation problems

PR #874, opened 2024-12-18
Invalidate cache

diPhantxm:


yacc/console/reserved_keyword.go
```diff
@@ -65,4 +65,6 @@ var reservedWords = map[string]int{
 	"redistribute": REDISTRIBUTE,
 	"batch":        BATCH,
 	"size":         SIZE,
+	"invalidate":   INVALIDATE,
```
Denchick, line 68: Don't forget to add why and how to use it in documentation

reshke (approved):

PR #880, opened 2024-12-20
Return correct logic to proxy routing

reshke:


Denchick:
Fixed here https://github.com/pg-sharding/spqr/pull/882

Denchick (approved):

Issue #897, opened 2024-12-26
Coordinator ListsRouters too often


Denchick:
I was investigating some incident, and found this problem in coordinator. Seems like Coordinator list router in etcd too often without any reason. This is how I see it in my logs:

```
{"level":"debug","time":"2024-12-26T12:27:45Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:46Z","message":"sending systemd notification: WATCHDOG=1\n"}
{"level":"debug","time":"2024-12-26T12:27:46Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:47Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:48Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:49Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:50Z","message":"sending systemd notification: WATCHDOG=1\n"}
{"level":"debug","time":"2024-12-26T12:27:50Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:51Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:53Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:54Z","message":"sending systemd notification: WATCHDOG=1\n"}
{"level":"debug","time":"2024-12-26T12:27:54Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:55Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:56Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:57Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:58Z","message":"sending systemd notification: WATCHDOG=1\n"}
{"level":"debug","time":"2024-12-26T12:27:58Z","message":"start routers watch iteration"}
{"level":"debug","time":"2024-12-26T12:27:59Z","message":"start routers watch iteration"}
{"level":"debug","client":824637126272,"message":{"Type":"Query","String":"SHOW version"},"time":"2024-12-26T12:28:00Z","message":"client received message"}
{"level":"debug","message":{"Type":"Query","String":"SHOW version"},"time":"2024-12-26T12:28:00Z","message":"received message"}
[...]

PR #903, opened 2024-12-27
Watch routers iteration timeout

Denchick:
Fixes https://github.com/pg-sharding/spqr/issues/897


coordinator/provider/coordinator.go
```diff
@@ -226,6 +228,9 @@ func (qc *qdbCoordinator) watchRouters(ctx context.Context) {
 			continue
 		}
 
+		// TODO we have to rewrite this code
```
reshke, line 231: use /**/

reshke (approved):

PR #945, opened 2025-01-15
Lower logging level of specific types of errors

reshke:


router/instance/instance.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
```
Denchick, line 5:
```suggestion
	"crypto/tls"
	"errors"
```

router/instance/instance.go
```diff
@@ -251,7 +252,11 @@ func (r *InstanceImpl) Run(ctx context.Context, listener net.Listener, pt port.R
 			} else {
 				go func() {
 					if id, err := r.serv(conn, pt); err != nil {
-						spqrlog.Zero.Error().Uint("client id", id).Err(err).Msg("error serving client")
+						if err == io.EOF || err == io.ErrUnexpectedEOF {
```
Denchick: if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {

Denchick (approved):

PR #1059, opened 2025-02-28
Simple role system (router only)

Denchick:


router/console/console.go
```diff
@@ -165,8 +171,24 @@ func (l *LocalInstanceConsole) Serve(ctx context.Context, cl client.Client) erro
...
 	}
-}
 
+}
 func (l *LocalInstanceConsole) Qlog() qlog.Qlog {
 	return l.qlogger
 }
+
+func (l *LocalInstanceConsole) checkGrants(target config.Role, rc rclient.RouterClient) error {
+	if !config.RouterConfig().EnableRoleSystem {
```
reshke, line 181: if g == admin then return immediately

reshke (approved):

PR #1075, opened 2025-03-12
Add REDISTRIBUTE CHECK command

EinKrebs:


pkg/datatransfers/data_transfers.go
```diff
@@ -236,24 +236,14 @@ func resolveNextBound(ctx context.Context, krg *kr.KeyRange, cr coordinator.Coor
...
-// - ds (*distributions.Distribution): the Distributions object representing the distribution of data.
-// - upperBound (kr.KeyRangeBound): the upper bound of the key range being moved.
-//
-// Returns:
-// - error: an error if the move fails.
-func copyData(ctx context.Context, from, to *pgx.Conn, fromId, toId string, krg *kr.KeyRange, ds *distributions.Distribution, upperBound kr.KeyRangeBound) error {
+func SetupFDW(ctx context.Context, from, to *pgx.Conn, fromId, toId string) error {
+	if shards == nil {
+		err := LoadConfig(config.CoordinatorConfig().ShardDataCfg)
+		if err != nil {
```
diPhantxm, line 242: Shouldn't we return error here? Otherwise we may panic further down

coordinator/provider/coordinator.go
```diff
@@ -1005,6 +1005,47 @@ func (qc *qdbCoordinator) Move(ctx context.Context, req *kr.MoveKeyRange) error
...
 
+// checkKeyRangeMove checks the ability to execute given kr.BatchMoveKeyRange request.
+//
+// Parameters:
+//   - ctx (context.Context): The context for QDB operations.
+//   - req (*kr.BatchMoveKeyRange): The move request to check.
+//
+// Returns:
+//   - error: An error if any occurred.
+func (qc *qdbCoordinator) checkKeyRangeMove(ctx context.Context, req *kr.BatchMoveKeyRange) error {
```
diPhantxm, line 1016: We may want to also check that relation exists on both shards

yacc/console/yx_test.go
```diff
@@ -283,6 +283,16 @@ func TestRedistribute(t *testing.T) {
 			},
 			err: nil,
 		},
+		{
+			query: "REDISTRIBUTE KEY RANGE kr1 TO sh2 CHECK",
```
Denchick, line 287: `CHECK REDISTRIBUTE KEY RANGE kr1 TO sh2` sounds better, doesn't it?

reshke: For manual execution, I think this is better for have this semantics: 

REDISTRIBUTE ... CHECK - dry run
REDISTRIBUTE ... APPLY - explicit request to start
REDISTRIBUTE ... - first check (abort if check failed), then start

EinKrebs: Can we do it in separate PR?

reshke (approved):

diPhantxm (approved):

PR #1121, opened 2025-04-14
Refactor base coordinator

diPhantxm:


reshke (approved):

test/regress/run_tests.sh
```diff
@@ -25,12 +25,30 @@ run_tests () {
...
+    done
+}
+
 
 run_tests "console" "regress_router" "7432"
 run_tests "router" "regress_router" "6432"
 run_tests "pooler" "regress_pooler" "6432"
 run_tests "coordinator" "regress_coordinator" "7002"
 
+run_tests "common" "regress_coordinator" "7002"
```
Denchick, line 48: Let's make a comment here? Smth like: these tests are to compare the results of the local and qdb coordinator.

Denchick (approved):

PR #1125, opened 2025-04-15
Add schema metadata to distributed relations

EinKrebs:
Also:
- Ablility to update relation metadata with `ALTER DISTRIBUTION ALTER RELATION` query
- Schema metadata is used in data transfers


reshke (approved):
LGTM

PR #1127, opened 2025-04-15
UUID

diPhantxm:


pkg/models/kr/keyrange.go
```diff
@@ -56,6 +56,10 @@ func (kr *KeyRange) InFuncSQL(attribInd int, raw []byte) {
 		fallthrough
 	case qdb.ColumnTypeVarchar:
 		kr.LowerBound[attribInd] = string(raw)
+	case qdb.ColumnTypeUUID:
+		val := string(raw)
+		val = strings.ToLower(val)
+		kr.LowerBound[attribInd] = val
```
EinKrebs, line 62: val can be inlined here

router/qrouter/proxy_routing.go
```diff
@@ -103,6 +103,7 @@ func (qr *ProxyQrouter) routingTuples(ctx context.Context, rm *rmeta.RoutingMeta
 	var p plan.Plan = nil
 
 	compositeKey := make([]interface{}, len(relation.DistributionKey))
+	spqrlog.Zero.Debug().Strs("col types", ds.ColTypes).Msg("here")
```
EinKrebs, line 106: Looks like dev logging, better remove

EinKrebs (changes requested):
Also tests should be added to yx_test.go

pkg/models/hashfunction/hashfunction.go
```diff
@@ -41,6 +43,13 @@ var (
 func ApplyHashFunction(inp interface{}, ctype string, hf HashFunctionType) (interface{}, error) {
 	switch hf {
 	case HashFunctionIdent:
+		if ctype == qdb.ColumnTypeUUID {
+			val := strings.ToLower(inp.(string))
+			err := uuid.Validate(val)
+			if err != nil {
+				return nil, err
+			}
```
EinKrebs, line 51:
```suggestion
			if err := uuid.Validate(strings.ToLower(inp.(string))); err != nil {
				return nil, err
			}
```
Can be inlined here

EinKrebs (changes requested):

diPhantxm:
> Also tests should be added to yx_test.go

I added uuid as column type in one of test cases, just to check it parses correctly

Denchick (approved):

EinKrebs (approved):

PR #1141, opened 2025-04-22
Fix multiple tuples on one distribution key in redistribute

EinKrebs:


coordinator/provider/coordinator.go
```diff
@@ -1367,7 +1369,10 @@ WHERE (sub.row_n %% constants.batch_size = 0 AND sub.row_n < constants.row_count
 		step,
 		rel.GetFullName(),
 		condition,
+		subColumns,
+		subColumns,
```
Denchick, line 1373: Is this typo or not?

Denchick (approved):

coordinator/provider/coordinator.go
Denchick: No

PR #1142, opened 2025-04-23
Restructure TSA checkers

Denchick:


pkg/pool/dbpool.go
```diff
@@ -27,7 +27,6 @@ type TsaKey struct {
 }
 
 type DBPool struct {
-	Pool
```
Denchick, line 30: Pool is an interface

reshke (approved):

PR #1143, opened 2025-04-23
IgnoreUnavailableHosts

Denchick:


pkg/pool/dbpool.go
```diff
@@ -293,11 +293,13 @@ func (s *DBPool) BuildHostOrder(key kr.ShardKey, targetSessionAttrs tsa.TSA) ([]
...
 		}
 
-		if res, ok := s.cacheTSAChecks.Load(tsaKey); ok {
+		if res, ok := s.CacheTSAChecks.Load(tsaKey); ok {
 			if res.(bool) {
 				posCache = append(posCache, host)
 			} else {
-				negCache = append(negCache, host)
+				if !config.RouterConfig().IgnoreUnavailableHosts {
+					negCache = append(negCache, host)
```
Denchick, line 301: Let's not just add `alive: false` hosts to the negCache because anyway they are unavailable.

Denchick:
Decided not to move forward with it

PR #1169, opened 2025-05-06
AUTO INCREMENT START syntax

Denchick:
The problem: if you already have a reference table on the shard, and you want to use the auto increment feature, then there is no way to set the initial value of the sequence. This PR adds yacc syntax for such an opportunity.


yacc/console/ast.go
```diff
@@ -106,6 +106,7 @@ type DistributionDefinition struct {
 	ColTypes             []string
 	Replicated           bool
 	AutoIncrementColumns []string
+	AutoIncrementStart   []string
```
diPhantxm, line 109: I would make `AutoIncrementColumns` a slice of some struct with `columnName` and `start` fields instead

yacc/console/gram.y
```diff
@@ -574,13 +576,23 @@ distribution_key_entry:
 	}
 
 distributed_relation_def:
-	RELATION any_id DISTRIBUTION KEY distribution_key_argument_list opt_auto_increment opt_schema_name
+	RELATION any_id DISTRIBUTION KEY distribution_key_argument_list opt_auto_increment opt_auto_increment_start opt_schema_name
```
diPhantxm, line 579: let's make `opt_auto_increment_start` part of `opt_auto_increment`? In this case we can specify START without actual AUTO INCREMENT

diPhantxm (approved):

reshke (approved):

Issue #803, opened 2024-11-11
default_route_behaviour: BLOCK 


Denchick:
If the default_route_behaviour is set to BLOCK, the router will send a message to the client:
```
ERROR:  client processing error: error processing query 'CREATE TABLE test_pers (puid INT, data JSONB);': failed to match any datashard., tx status IDLE
```
This message can be misleading for users, so it would be better if the router could answer with a more informative message, such as "Your query is blocked due to the default route behavior."

Firsikita:
Hello, I think I made a mistake. Here is the PR for this task: https://github.com/pg-sharding/spqr/pull/1907

PR #1172, opened 2025-05-07
Specify message when query is blocked due to the default route behavior

PavelLovygin:
…e behavior


Denchick:
Hi there! 
Thank you for your contribution to our project!
It seems like your PR fixes #803. 

Could you please provide some tests for this PR? For instance, [in the regress tests](https://github.com/pg-sharding/spqr/blob/master/test/regress/tests/router/sql/routing_hint.sql#L91C72-L91C103), you could explicitly set the default route behavior to BLOCK and check if you actually get an error with the specified message.

Smth like: `CREATE TABLE denchick(i int) /* __spqr__default_route_behaviour: block */;`

PR #1180, opened 2025-05-12
Add reset support for __spqr__ tsa param

reshke:


Denchick (approved):
Could you please add tests and documentation?

PR #1181, opened 2025-05-12
Fix for pure virtual processing

reshke:


router/plan/plan.go
```diff
@@ -131,6 +131,12 @@ func Combine(p1, p2 Plan) Plan {
 		Interface("plan2", p2).
 		Msg("combine two plans")
 
+	switch p2.(type) {
+	// let p2 be always non-virtual, expect for p1 & p2 both virtual
+	case VirtualPlan:
+		p1, p2 = p2, p1
```
Denchick, line 137: I have no idea what happening here. What deduces mean, why smth is questionable?... anyway, seems like you should add comments

Denchick (approved):

PR #1258, opened 2025-06-11
fix intersection key ranges

xelavopelk:
incorrect check of intersection for key_range which lets create doubles
example:
3 runs init2shards.sql gets memqdb_with_doubles.json  
[init2shards.txt](https://github.com/user-attachments/files/20692054/init2shards.txt)
[memqdb_with_doubles.json](https://github.com/user-attachments/files/20692055/memqdb_with_doubles.json)


xelavopelk:
not correct

Issue #1207, opened 2025-05-21
show backend_connections group by fails;


reshke:
spqr-console=> show backend_connections group by host;
ERROR:  syntax error
spqr-console=> show backend_connections group by hostnae;
SSL SYSCALL error: EOF detected
The connection to the server was lost. Attempting reset: Failed.
The connection to the server was lost. Attempting reset: Failed.
!?> show backend_connections group by hostnae;^C
!?>

PR #1273, opened 2025-06-16
fix issue 1207, refactor groupby

xelavopelk:
0) fix issue 1207
1) group by method is refactored to be less fragile


reshke:
Looks like PR missing changes to .gitignore
also, how about checking regression test from #1207 ?

reshke (approved):

PR #1287, opened 2025-06-24
MVP default shards

xelavopelk:
Default shards MVP
The main idea of default shard realization is creating key range with minimal value for distribution's column set.
Supported commands:
- create distribution with default shard: CREATE DISTRIBUTION testWithDefault COLUMN TYPES integer DEFAULT SHARD sh2;
- drop default shard of distribution: ALTER DISTRIBUTION testWithDefault DROP DEFAULT SHARD;
- add default shard to existed distribution: ALTER DISTRIBUTION testWithDefault DEFAULT SHARD sh2;

### DIFF
--- a/pkg/coord/clocal.go
+++ b/pkg/coord/clocal.go
@@ -340,7 +340,15 @@ func (lc *LocalInstanceMetadataMgr) SyncRouterCoordinatorAddress(ctx context.Con
 // - *topology.DataShard: The retrieved DataShard, or nil if it doesn't exist.
 // - error: An error indicating the retrieval status, or ErrNotCoordinator if the operation is not supported by the LocalCoordinator.
 func (lc *LocalInstanceMetadataMgr) GetShard(ctx context.Context, shardID string) (*topology.DataShard, error) {
-	return nil, ErrNotCoordinator
+	if lc.qdb != nil {
+		sh, err := lc.qdb.GetShard(ctx, shardID)
+		if err != nil {
+			return nil, err
+		}
+		return topology.DataShardFromDB(sh), nil
+	} else {
+		return nil, ErrNotCoordinator
+	}
 }
 
 func (lc *LocalInstanceMetadataMgr) Cache() *cache.SchemaCache {
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -2,6 +2,7 @@ package meta
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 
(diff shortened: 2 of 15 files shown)


pkg/coord/clocal.go
```diff
@@ -340,7 +340,15 @@ func (lc *LocalInstanceMetadataMgr) SyncRouterCoordinatorAddress(ctx context.Con
...
 // - error: An error indicating the retrieval status, or ErrNotCoordinator if the operation is not supported by the LocalCoordinator.
 func (lc *LocalInstanceMetadataMgr) GetShard(ctx context.Context, shardID string) (*topology.DataShard, error) {
-	return nil, ErrNotCoordinator
+	if lc.qdb != nil {
+		sh, err := lc.qdb.GetShard(ctx, shardID)
+		if err != nil {
+			return nil, err
+		}
+		return topology.DataShardFromDB(sh), nil
+	} else {
```
Denchick: I found this if-else "problem" in many places in your PR. It is more about style and readability than logic. 

In case you are doing something like this:

```
func important() error {
    if err != nil {
        return err
    } else {
        return nil
    }
}
```

you don't need the `else` keyword:

```
func important() error {
    if err != nil {
        return err
    }
    return nil
}
```

pkg/meta/meta.go
```diff
@@ -346,6 +388,26 @@ func processAlterDistribution(ctx context.Context, astmt spqrparser.Statement, m
 			return err
 		}
 		return cli.AlterDistributedRelation(ctx, stmt.Distribution.ID, stmt.Relation.Name)
+	case *spqrparser.DropDefaultShard:
```
Denchick, line 412: I feel kinda confused because `TrySmth` isn't a golang-way for me...
If I were writing this code, I would write it like this:

```golang
 	case *spqrparser.DropDefaultShard:
		if distribution, err := mngr.GetDistribution(ctx, distributionId); err != nil {
			return nil, err
		}
		if defaultShard, err := manager.DropDefaultShard(ctx); err != nil {
			return err
		}
		return cli.MakeSimpleResponse(ctx, manager.SuccessDropResponse(*defaultShard))
	case *spqrparser.AlterDefaultShard:
		if distribution, err := mngr.GetDistribution(ctx, distributionId); err != nil {
			return nil, err
		}
		if err := manager.CreateDefaultShard(ctx, stmt.Shard); err != nil {
			return err
		}
		return cli.MakeSimpleResponse(ctx, manager.SuccessCreateResponse(stmt.Shard))
```

reshke (approved):

Denchick (changes requested):
First of all, I want to thank you very much for your PR. It is really valuable and I'm glad that we now have the default shard feature. 

As it seems to me, there were some minor style-related comments left by me, but the most important thing I wanted to add was that we need to document this feature and how to use it. I think this can be done through a separate PR.

Do you need help with documentation? We need to:
- Add commands to [the SQL reference](https://github.com/pg-sharding/spqr/blob/master/docs/sharding/console/sql_commands.mdx).
- Add the default shard feature to [the list](https://github.com/pg-sharding/spqr/tree/master/docs/features).
- Write an article-guide about default sharding [here](https://github.com/pg-sharding/spqr/tree/master/docs/sharding).

pkg/coord/clocal.go
xelavopelk: fixed except cases when i need variable in "if" scope. like that. it's ok?
```go	
if defaultShard, err := manager.mngr.GetShard(ctx, defaultShardId); err != nil {
		return fmt.Errorf("shard '%s' for default is not exists", defaultShard.ID)
	} else {
		return manager.CreateDefaultShardNoCheck(ctx, defaultShard)
	}
```

pkg/meta/meta.go
xelavopelk: fixed

xelavopelk:
docs https://github.com/pg-sharding/spqr/pull/1291

Denchick (approved):

PR #1297, opened 2025-06-29
Add regression tests for `default shard` feature.

reshke:
Also adjust console sql for default shard definition: expect `ADD` in `ALTER DISTRIBUTION`


Denchick (approved):

xelavopelk:
need change list of supported types "Default shards are supporting only `integer` and `varchar` types now." in default_shards.mdx

PR #1299, opened 2025-07-01
Add basic feature test for reference relation

reshke:


test/feature/features/reference_relation.feature
```diff
@@ -0,0 +1,94 @@
+Feature: Sequence test
```
Denchick: Typo

PR #1320, opened 2025-07-04
Sync reference relation data transfer utilities

reshke:


test/regress/tests/console/expected/reference_table.out
```diff
@@ -22,7 +22,7 @@ CREATE REFERENCE TABLE zz2 ON sh3, sh4;
 CREATE REFERENCE RELATION zz2;
 ERROR:  reference relation zz2 already exists
 SYNC REFERENCE TABLES ON sh1;
-ERROR:  sync reference table cmd not implemented
+ERROR:  request is unprocessable in router
```
Denchick, line 25: What should i do? _connect to a coordiantor or infra host and try again_

pkg/meta/meta.go
```diff
@@ -576,7 +576,11 @@ func ProcMetadataCommand(ctx context.Context, tstmt spqrparser.Statement, mgr En
 		}
 		return cli.MoveTaskGroup(ctx, taskGroup, colTypes)
 	case *spqrparser.SyncReferenceTables:
-		return cli.ReportError(fmt.Errorf("sync reference table cmd not implemented"))
+		/* todo: fix */
```
Denchick: explain what we need to fix here?

Denchick (approved):

test/regress/tests/console/expected/reference_table.out
reshke: yes, but this is subject to separate discussion, because this error message is not specific to this command

PR #1337, opened 2025-07-07
display host alive status in `SHOW hosts`

reshke:


rkhapov (approved):

pkg/tsa/cachedchecker.go
```diff
@@ -14,6 +14,13 @@ type CachedTSACheckerImpl struct {
 	innerChecker  TSAChecker
 }
 
+// InstanceHealthChecks implements CachedTSAChecker.
+func (ctsa *CachedTSACheckerImpl) InstanceHealthChecks() map[string]CachedCheckResult {
+	ctsa.mu.Lock()
+	defer ctsa.mu.Unlock()
+	return ctsa.cache
```
Denchick: if you want to safely pass underlying map, you have to copy it. any other goroutine could modify the map after the lock is released

PR #1342, opened 2025-07-07
fix uint and int encoding

xelavopelk:


pkg/models/hashfunction/hashfunction.go
```diff
@@ -97,7 +97,7 @@ func ApplyHashFunction(input interface{}, ctype string, hf HashFunctionType) (in
 	case HashFunctionCity:
 		switch ctype {
 		case qdb.ColumnTypeInteger:
-			buf := make([]byte, 8)
+			buf := make([]byte, binary.MaxVarintLen64)
```
reshke: lets change this to hashfunction.EncodeUInt64(number) too

reshke (approved):

PR #1358, opened 2025-07-09
Fix sync reference relation storage shards metadata

reshke:


pkg/coord/adapter.go
```diff
@@ -74,7 +75,26 @@ func (a *Adapter) GetReferenceRelation(ctx context.Context, relName *rfqn.Relati
...
+	_, err := c.SyncReferenceRelations(ctx, &proto.SyncReferenceRelationsRequest{
+		Relations: qRels,
+		ShardId:   destShard,
+	})
+	return err
+}
+
+// AlterReferenceRelationStorage implements meta.EntityMgr.
+func (a *Adapter) AlterReferenceRelationStorage(ctx context.Context, relName *rfqn.RelationFQN, shs []string) error {
+	return fmt.Errorf("AlterReferenceRelationStorage should not be used in proxy adapter")
```
Denchick, line 97: I don't understand, why not?

Denchick (changes requested):
Agreed with minor comment of copilot

pkg/coord/adapter.go
reshke: because this is coordinator -> router type of message only

Issue #983, opened 2025-01-29
Do not ignore schema of a table


Denchick:
When you are adding a new table, you can specify a schema. For example:

```
ALTER DISTRIBUTION ds1 ATTACH RELATION schema_name.table_name DISTRIBUTION KEY id;
```

In fact, during query/transaction routing, the schema is not taken into account. If there are tables with the same name but different schemas, they will be incorrectly routed. To fix this bug, you have to remember about `search_path` as well.

xelavopelk:
"RELATION any_id DISTRIBUTION KEY distribution_key_argument_list opt_auto_increment opt_schema_name
...
opt_schema_name:
	SCHEMA any_id {
		$$ = $2
	} | /* EMPTY */ {
		$$ = ""
	}
"
schema sets in schema clause. May be it's not bug, but a language feature. Does we need setup schema by 2 ways?

"ALTER DISTRIBUTION ds1 ATTACH RELATION schema_name.table_name DISTRIBUTION KEY id;"
equals
"ALTER DISTRIBUTION ds1 ATTACH RELATION table_name DISTRIBUTION KEY id SCHEMA schema_name;"

is it right?

Denchick:
Hi!

While both syntaxes are valid, the dot notation is more SQL-way, i think.

Deprecating the SCHEMA keyword version would simplify things and align with common practice. What do you think?

Denchick:
Fixed in https://github.com/pg-sharding/spqr/pull/1298.
@xelavopelk thank you for your contribution!

Denchick:
No, #1298 was only the beginning of fixing problems.

Denchick:
https://github.com/pg-sharding/spqr/pull/1326

Denchick:
We paused https://github.com/pg-sharding/spqr/pull/1366 because the changes were becoming too dangerous for the existing installation. We had no intention of breaking backward compatibility.

Denchick:
Stale

PR #1366, opened 2025-07-10
issue 983. change internal api for schema using

xelavopelk:
Distribution.Relations is really internal api too
routing bhaviour changed: if the relation created without schema we can't ask it using name with a shema.
example:
create table t1 (id int)
...
select * from sch.t1 >>> error
select * from t1 >> ok
PS default "public" schema is not supported yet


router/qrouter/proxy_routing_test.go
```diff
@@ -858,17 +858,6 @@ func TestSingleShard(t *testing.T) {
 			},
 			err: nil,
 		},
-		{
-			query: "SELECT * FROM sh1.xxtt1 WHERE sh1.xxtt1.i = 21;",
```
Denchick, line 880: Why did you remove this tests?

Denchick:
We decided to pause this PR. Please do not close.
https://github.com/pg-sharding/spqr/issues/983#issuecomment-3178405414

PR #1398, opened 2025-07-17
Support __spqr__.* virtual param names

reshke:


test/regress/tests/router/expected/virtual_show.out
```diff
@@ -97,3 +103,17 @@ SHOW __spqr__engine_v2;
  ok
 (1 row)
 
+SET __spqr__.engine_v2 TO OFF;
+SHOW __spqr__.engine_v2;
```
reshke, line 107: no is alias for off in internal logic

PR #1400, opened 2025-07-17
persistent store sequence value after increment

xelavopelk:
persistent store in memDB sequence value after increment


qdb/memqdb.go
```diff
@@ -1114,7 +1114,11 @@ func (q *MemQDB) NextVal(_ context.Context, seqName string) (int64, error) {
 
 	next := q.SequenceToValues[seqName] + 1
 	q.SequenceToValues[seqName] = next
-	return next, nil
+	if errDB := ExecuteCommands(q.DumpState, NewUpdateCommand(q.SequenceToValues, seqName, next)); errDB != nil {
+		return next, errDB
+	} else {
```
Denchick: You can make it a bit shorter
```
if errDB := ExecuteCommands(q.DumpState, NewUpdateCommand(q.SequenceToValues, seqName, next)); errDB != nil {
	return next, errDB
}
return next, nil
```

xelavopelk: fixed

PR #1406, opened 2025-07-18
show users

diPhantxm:


reshke (approved):
dont forget to add regression test to test/regress/tests/console/something

PR #1439, opened 2025-07-25
Small typo fix

Pum-purum:


reshke:
I wonder how its used to work 
copy-pasted from here
https://github.com/pg-sharding/cpg/blob/MDB_13_18/docker/entrypoint.sh#L4
Maybe we should just delete this and use UTC+0?

reshke:
regression test didnt like this change 
```
 NOTICE: send query to shard(s) : sh1
       to_timestamp      
 ------------------------
- 2017-03-31 09:30:20+00
+ 2017-03-31 09:30:20+03
 (1 row)
 
```

reshke:
Merged, thanks!

PR #1444, opened 2025-07-26
Enhance SPQR sharding DDL

reshke:
Now we can omit distribution definition in case there is only one


test/regress/tests/console/expected/default_distribution.out
```diff
@@ -40,6 +47,8 @@ CREATE DISTRIBUTION ds2 COLUMN TYPES integer;
 
 CREATE KEY RANGE krid3 FROM 11 ROUTE TO sh2;
 ERROR:  distributions count not equal one, use FOR DISTRIBUTION syntax
+CREATE DISTRIBUTED RELATION r1 (id);
+ERROR:  distributions count not equal one, use FOR DISTRIBUTION syntax
```
Denchick, line 51: There is more than one distribution, use bla bla bla

PR #1452, opened 2025-07-28
Issue 1441. MVP

xelavopelk:
MVP.  Router get the range for generationg id. Range size is setted up in a router config. This range size is default for all sequences.


reshke (approved):
Actually looks good

pkg/config/router.go
```diff
@@ -100,6 +100,7 @@ type Router struct {
 
 	UseSystemdNotifier   bool `json:"use_systemd_notifier" toml:"use_systemd_notifier" yaml:"use_systemd_notifier"`
 	SystemdNotifierDebug bool `json:"systemd_notifier_debug" toml:"systemd_notifier_debug" yaml:"systemd_notifier_debug"`
+	IdentityRangeSize    int  `json:"identity_range_size" toml:"identity_range_size" yaml:"identity_range_size"`
```
Denchick: Add this setting to the docs/ please.

Denchick: Why is this value not uint64?

router/planner/identity_router_cache.go
```diff
@@ -0,0 +1,92 @@
...
+
+type CachedIdRange struct {
+	idRange *qdb.SequenceIdRange
+	mu      sync.Mutex
+}
+
+func NewCachedIdRange() CachedIdRange {
+	return CachedIdRange{mu: sync.Mutex{}}
+}
+func (cir *CachedIdRange) nextVal() (int64, bool) {
```
Denchick, line 26:
```
func (cir *CachedIdRange) nextVal() (int64, bool) {
	if cir.idRange == nil {
		return 0, false
	}
	if cir.idRange.Left < cir.idRange.Right {
		res := cir.idRange.Left
		cir.idRange.Left++
		return res, true
	}
	if cir.idRange.Left == cir.idRange.Right {
		res := cir.idRange.Left
		cir.idRange = nil //the range is over
		return res, true
	}
	return 0, false
}
```

PR #1492, opened 2025-08-05
fix: shards healthcheck (docker-compose sample)

nryanov:
Hi!

It seems there was a typo. Postgres shards listening port 6543, but pg_isready expecting by default 5432. Because of that shards always in state `unhealthy` in docker. With this fix containers eventually become `healthy`


Denchick (approved):
LGTM

PR #1518, opened 2025-08-20
Manage shards metadata via coordinator

diPhantxm:


test/feature/features/default_shard.feature
```diff
@@ -12,8 +12,6 @@ Feature: default shards test
     REGISTER ROUTER r1 ADDRESS regress_router:7000;
     CREATE DISTRIBUTION ds1 COLUMN TYPES integer;
     ALTER DISTRIBUTION ds1 ATTACH RELATION xMove DISTRIBUTION KEY w_id;
-    ADD SHARD sh1 WITH HOSTS 'postgresql://regress@spqr_shard_1:6432/regress';
```
Denchick: i don't quite understand why are you deleting ADD SHARD commands

Denchick (changes requested):
Maybe we need some feature flag how to manage shards. Smth like ManageShardsVia:file|etcd flag in the router config

Because I don't quite understand how to smoothly organise the transition from file-based management to etcd management.

test/feature/features/default_shard.feature
diPhantxm: Because we actually have already added them [here](https://github.com/pg-sharding/spqr/blob/master/pkg/coord/clustered_coord.go#L334)
But since we don't have a check if shard already exists, when adding shard to qdb in clustered coordinator ([here](https://github.com/pg-sharding/spqr/blob/master/qdb/etcdqdb.go#L825)), we didn't receive any error. But now we send request to each router to add this shard to their memqdb, and they have this check ([here](https://github.com/pg-sharding/spqr/blob/master/qdb/memqdb.go#L578))

diPhantxm:
> Maybe we need some feature flag how to manage shards. Smth like ManageShardsVia:file|etcd flag in the router config
> 
> Because I don't quite understand how to smoothly organise the transition from file-based management to etcd management.

Well, I don't know either. It may be confusing for some users who have already setup shards in config files and it works fine, they won't expect any changes here. But I believe this implementation is how it should have been done from the very beginning.

xelavopelk:
I propose following design:
1) We'll use shard configuration in config file as feature flag. We use shard configuration if config file contains any shard. In this mode we can't use "ADD/DROP SHARD" commands. Old clients in this mode will work us usual. It will be named "fixed mode".
When no shard is in config file we'll use database as source of shard configuration.  "ADD/DROP SHARD" commands will work at this mode. This mode should be preferred for new installations. This should be indicated in the documentation. It will be named "flex mode".
2) We must check that shard is not used before drop shard. Database may became to inconsistent state without this check.

diPhantxm:
> I propose following design:
> 
> 1. We'll use shard configuration in config file as feature flag. We use shard configuration if config file contains any shard. In this mode we can't use "ADD/DROP SHARD" commands. Old clients in this mode will work us usual. It will be named "fixed mode".
>    When no shard is in config file we'll use database as source of shard configuration.  "ADD/DROP SHARD" commands will work at this mode. This mode should be preferred for new installations. This should be indicated in the documentation. It will be named "flex mode".
> 2. We must check that shard is not used before drop shard. Database may became to inconsistent state without this check.

Sure, thank you for your proposal. I will return and finish this pr asap as you mentioned it as blocker.

test/feature/features/coordinator.feature
```diff
@@ -464,8 +464,7 @@ Feature: Coordinator test
   Scenario: Adding/dropping shards works
     When I run SQL on host "coordinator"
     """
-    ADD SHARD sh3 WITH HOSTS "spqr_shard_1:6432";
```
Denchick, line 467: what was happening here before? I just don't understand why did you update this test

router/instance/etcd.go
```diff
@@ -28,6 +30,32 @@ func (e *EtcdMetadataBootstrapper) InitializeMetadata(ctx context.Context, r Rou
 		}
 	}()
 
+	/* Initialize shards */
+	if config.RouterConfig().ManagedShardsByCoordinator {
```
Denchick: For unittesting purpuses i guess it is better to pass this variable as a parameter

pkg/config/router.go
```diff
@@ -72,10 +72,11 @@ type Router struct {
...
 
-	InitSQL            string `json:"init_sql" toml:"init_sql" yaml:"init_sql"`
-	UseInitSQL         bool   `json:"use_init_sql" toml:"use_init_sql" yaml:"use_init_sql"`
-	ExitOnInitSQLError bool   `json:"exit_on_init_sql" toml:"exit_on_init_sql" yaml:"exit_on_init_sql"`
-	UseCoordinatorInit bool   `json:"use_coordinator_init" toml:"use_coordinator_init" yaml:"use_coordinator_init"`
+	InitSQL                    string `json:"init_sql" toml:"init_sql" yaml:"init_sql"`
+	UseInitSQL                 bool   `json:"use_init_sql" toml:"use_init_sql" yaml:"use_init_sql"`
+	ExitOnInitSQLError         bool   `json:"exit_on_init_sql" toml:"exit_on_init_sql" yaml:"exit_on_init_sql"`
+	UseCoordinatorInit         bool   `json:"use_coordinator_init" toml:"use_coordinator_init" yaml:"use_coordinator_init"`
+	ManagedShardsByCoordinator bool   `json:"manage_shards_by_coordinator" yaml:"manage_shards_by_coordinator"`
```
Denchick: Typo? Managed vs. manage_
Btw, where is toml?

pkg/coord/clocal.go
```diff
@@ -193,10 +195,17 @@ func (lc *LocalInstanceMetadataMgr) AddDataShard(ctx context.Context, ds *topolo
 		Str("node", ds.ID).
 		Msg("adding datashard node in local coordinator")
 
-	return lc.qdb.AddShard(ctx, &qdb.Shard{
-		ID:       ds.ID,
-		RawHosts: ds.Cfg.RawHosts,
-	})
+	if config.RouterConfig().ManagedShardsByCoordinator {
```
Denchick: It is better to pass the config setting to NewLocalInstanceMetadataMgr as a variable.

coordinator/pkg/clustered_coord.go
```diff
@@ -1870,6 +1870,32 @@ func (qc *ClusteredCoordinator) SyncRouterMetadata(ctx context.Context, qRouter
...
+	shards, err := qc.ListShards(ctx)
+	if err != nil {
+		return err
+	}
+	shardResp, err := shCl.ListShards(ctx, nil)
+	if err != nil {
+		return err
+	}
+	for _, sh := range shardResp.GetShards() {
+		_, err = shCl.DropShard(ctx, &proto.DropShardRequest{
```
Denchick: Why do you need recreate shards every time when somebody want to sync it?
What if there are key ranges assigned to this shard?

pkg/coord/clocal.go
```diff
@@ -193,10 +195,17 @@ func (lc *LocalInstanceMetadataMgr) AddDataShard(ctx context.Context, ds *topolo
...
+		lc.shardMapping[ds.ID] = ds.Cfg
+	}
+	return lc.Coordinator.AddDataShard(ctx, ds)
+}
+
+func (lc *LocalInstanceMetadataMgr) DropShard(ctx context.Context, shardId string) error {
+	if config.RouterConfig().ManagedShardsByCoordinator {
+		delete(lc.shardMapping, shardId)
```
Denchick, line 218: You should lock the map before modifying it

coordinator/pkg/clustered_coord.go
```diff
@@ -1870,6 +1870,32 @@ func (qc *ClusteredCoordinator) SyncRouterMetadata(ctx context.Context, qRouter
...
 	}
 
+	// Configure shards
+	shCl := proto.NewShardServiceClient(cc)
+	spqrlog.Zero.Debug().Msg("qdb coordinator: configure shards")
+	shards, err := qc.ListShards(ctx)
+	if err != nil {
+		return err
+	}
+	shardResp, err := shCl.ListShards(ctx, nil)
```
Denchick:
```suggestion
	shardResp, err := shCl.ListShards(ctx, &emptypb.Empty{})
```

coordinator/pkg/clustered_coord.go
diPhantxm: Because this is what we do with all other stuff here. We delete distributions, then create new ones, those from coordinator. The same for key ranges. I think it's just easier to recreate than compare and adjust. Like if address of shard has changed, we don't change the address of the existing shard, we just recreate it

test/feature/features/coordinator.feature
diPhantxm: Well, there was a mistake adding shard sh3, but specify address of shard 1. I encountered some unexpected behavior of connection pool. Now it's shard3 with name sh3, but I removed shard 4 and I don't remember why, but we can return shard 4 but with the correct address.
I also extended this test, to verify new shards exist on router and it can send queries on them.

diPhantxm:
> It is no clear for me, how to safely migrate from file-based approach to etcd-based approach?

I believe you should add all the shards on coordinator, so they appear in etcd and can be listed via `show command`, then you register your routers if not already, and then set `manage_shards_by_coordinator` in router config and restart. During sync or etcd initialization all shards will be added from etcd

test/feature/features/coordinator.feature
diPhantxm: I've just checked there is no shard4 in feature tests, so I think it's better to leave it as it is in this PR. Or add a new shard to docker compose in feature tests, but I don't find benefits from it

router/instance/etcd.go
diPhantxm: If we want to use function parameter for this, we would need to modify interface, but this flag is only used in ETCD bootstrapper, sqlinit does not need it. I'm not sure it's a good idea to modify interface and add something that will be used by only one implementation

Denchick:
@reshke could you review this PR as well?

coordinator/pkg/clustered_coord.go
Denchick: I understand the consistency argument, but I think we have an opportunity to do better here and avoid a potential issue.

  The concern: What happens if a router receives a query between the drop and add operations? We could see errors or latency spikes during the sync window — and in the worst case, this might trigger cascading failures.

  Suggested approach: Instead of drop-all/add-all, we could do a diff-based sync:

  1. Block shard changes (this should be possible after @xelavopelk's changes land)
  2. Compare desired shard state vs actual state
  3. Only drop removed shards, add missing ones, and leave unchanged shards alone

  This would make the sync operation safer with minimal additional complexity. What do you think?

Denchick:
As part of this PR, we forgot to think about how it would work PreferSameAvailabilityZone feature https://github.com/pg-sharding/spqr/issues/1856

xelavopelk:
may be needs split task to small parts?

diPhantxm:
> may be needs split task to small parts?

Ahhh, I think this task is done, at least I've written the code, it seems to be working. Just need a review. All previous issues are fixed

PR #1544, opened 2025-09-08
Recover dead hosts in dbpool

Denchick:


pkg/pool/dbpool.go
```diff
@@ -39,6 +40,179 @@ type DBPool struct {
...
+
+	switch requiredTSA {
+	case config.TargetSessionAttrsRW:
+		return cr.RW
+	case config.TargetSessionAttrsSmartRW:
+		return cr.RW
+	case config.TargetSessionAttrsRO:
+		return !cr.RW
+	case config.TargetSessionAttrsPS:
+		return !cr.RW // prefer standby (read-only)
```
Denchick, line 201: TBD: whether we should return a score instead of a true/false value.

PR #1547, opened 2025-09-09
Add etcd max request size param to coordinator config

EinKrebs:


xelavopelk:
Add it into documentation

PR #1558, opened 2025-09-15
expression_solver

xelavopelk:
proof of concept.
solve problem with range of sharded keys.
example:

select * from Comments where hashslot>0 and hashslot < 8192;
select * from Comments where hashslot < 5 and hashslot > 8000;


xelavopelk:
does not match the architecture

PR #1573, opened 2025-09-21
fix locks

xelavopelk:
0) LockKeyRange has been made threadsafe and more consistent (by checking for the existence of the lock and key range in transaction).
 1) I propose FastLockKeyRange qdb function. I propose this function for using in Split, Unite, RenameKeyRange and others which make changes ONLY in qdb. PROPOSED BEHAVIOUR for them: If lock fails at the start of that functions then function fails fast without redundant waits. If anybody wants any waits he have to make it in business-logic level. If ok I'll use FastLockKeyRange in that funct

### DIFF
--- a/qdb/etcdqdb.go
+++ b/qdb/etcdqdb.go
@@ -340,38 +340,95 @@ func (q *EtcdQDB) ListAllKeyRanges(ctx context.Context) ([]*KeyRange, error) {
 	return ret, nil
 }
 
-// TODO : unit tests
-func (q *EtcdQDB) LockKeyRange(ctx context.Context, id string) (*KeyRange, error) {
+func (q *EtcdQDB) FastLockKeyRange(ctx context.Context, idKeyRange string) (*KeyRange, error) {
 	spqrlog.Zero.Debug().
-		Str("id", id).
+		Str("id", idKeyRange).
+		Msg("etcdqdb: lock key range (fast)")
+	t := time.Now()
+	kr, err := q.internalFastLockKeyRange(ctx, idKeyRange)
+	statistics.RecordQDBOperation("LockKeyRange", time.Since(t))
+	return kr, err
+}
+
+func (q *EtcdQDB) internalFastLockKeyRange(ctx context.Context, idKeyRange string) (*KeyRange, error) {
+
+	resp, err := q.cli.Txn(ctx).
+		If(
+			//check exists key range lock
+			clientv3.Compare(clientv3.Version(keyLockPath(keyRangeNodePath(idKeyRange))), "=", 0),
+			//check exists key range
+			clientv3.Compare(clientv3.Version(keyRangeNodePath(idKeyRange)), ">", 0),
+		).
+		Then(
+			clientv3.OpPut(keyLockPath(keyRangeNodePath(idKeyRange)), "locked"),
+			clientv3.OpGet(keyRangeNodePath(idKeyRange)),
+		).
+		Else(
+			clientv3.OpGet(keyLockPath(keyRangeNodePath(idKeyRange)), clientv3.WithCountOnly()),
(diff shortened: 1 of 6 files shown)


qdb/etcdqdb.go
```diff
@@ -340,38 +340,95 @@ func (q *EtcdQDB) ListAllKeyRanges(ctx context.Context) ([]*KeyRange, error) {
 	return ret, nil
 }
 
-// TODO : unit tests
-func (q *EtcdQDB) LockKeyRange(ctx context.Context, id string) (*KeyRange, error) {
+func (q *EtcdQDB) FastLockKeyRange(ctx context.Context, idKeyRange string) (*KeyRange, error) {
```
Denchick: The keyword "fast" does not reveal the main difference from the original LockKeyRange method. Is it really more suitable to use the name NOWAIT?

test/feature/features/coordinator.feature
```diff
@@ -359,6 +359,16 @@ Feature: Coordinator test
     context deadline exceeded
     """
 
+    When I run SQL on host "coordinator"
+    """
+    LOCK KEY RANGE krid1NoExistS;
+    """
+    Then SQL error on host "coordinator" should match regexp
+    """
+    cant't lock non existent key range
```
Denchick, line 368: Seems like you have to add a testcase here when fastlockkeyrange fails here because it can't wait smth?

qdb/etcdqdb.go
```diff
@@ -340,38 +340,95 @@ func (q *EtcdQDB) ListAllKeyRanges(ctx context.Context) ([]*KeyRange, error) {
...
+}
+
+// TODO : unit tests
+func (q *EtcdQDB) LockKeyRange(ctx context.Context, idKeyRange string) (*KeyRange, error) {
+	spqrlog.Zero.Debug().
+		Str("id", idKeyRange).
+		Msg("etcdqdb: lock key range")
+
+	t := time.Now()
+	if kr, err := retry.DoValue(ctx, retry.WithMaxRetries(7,
```
Denchick: Does it make any sense to configure these values? Or maybe just move them to constants?

qdb/etcdqdb.go
```diff
@@ -340,38 +340,95 @@ func (q *EtcdQDB) ListAllKeyRanges(ctx context.Context) ([]*KeyRange, error) {
...
-			return retry.RetryableError(err)
+	kr, err := q.internalFastLockKeyRange(ctx, idKeyRange)
+	statistics.RecordQDBOperation("LockKeyRange", time.Since(t))
+	return kr, err
+}
+
+func (q *EtcdQDB) internalFastLockKeyRange(ctx context.Context, idKeyRange string) (*KeyRange, error) {
+
+	resp, err := q.cli.Txn(ctx).
+		If(
```
Denchick, line 358: I don't like this versioning logic because it feels like everything is thread-safe now, but it's still not. Don't mislead other developers.

xelavopelk: not everithing. Only lock. in this PR check lock and put lock is one single atomic operation.  Now two or more threads can get a lock of key_range. This is not possible in the proposed version.

qdb/etcdqdb.go
xelavopelk: +

qdb/etcdqdb.go
xelavopelk: +

test/feature/features/coordinator.feature
xelavopelk: fastlockkeyrange (now NoWaitLockKeyRange as you propose) haven't any logic. It's just api wrapper over internalNoWaitLockKeyRange. internalNoWaitLockKeyRange is testsed here in fact.

Issue #1458, opened 2025-07-29
Correctly print running router config on start when overriding values from CLI options


reshke:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Denchick:
**Current behavior**:
When starting the router with `-c config.yaml` and overriding some values via CLI options, the “Running config” printed on startup does not reflect the overridden values — it still shows the original values from the YAML file.

Example:

```bash
./spqr-router run -c test/stress/router.yaml --log-level=info
2025/08/12 10:15:49 Running config: {
  "log_level": "fatal",
...
```

**Expected behavior:**

On startup, the printed “Running config” should include all effective values, including those overridden via CLI options.

After a SIGHUP signal (e.g., service spqr-router reload), CLI-overridden values should remain in the effective config and continue to be reflected when the config is printed.

```bash
./spqr-router run -c test/stress/router.yaml --log-level=info
2025/08/12 10:15:49 Running config: {
  "log_level": "info",
...
```

futig:
I take it

Denchick:
Hi @futig ! How is your progress going? We've seen your https://github.com/pg-sharding/spqr/pull/1582, and we're curious to know whether you've already debugged why the tests are failing. Maybe you need some help? Feel free to ask!

futig:
Sorry, my bad, had some troubles with tests. I decided to close #1582  PR and create the new one #1585

Denchick:
@futig many thanks!

PR #1585, opened 2025-10-02
Task #1458 cli override

futig:


reshke:
I wonder if we can actually test this as least in a few simple cases 

maybe something like cli_test.sh with example configs and cli options will do?

futig:
Added tests to show cli override

futig:
I updated expected words in spelling (stdbuf + drb) and merged upstream master

futig:
Updated pipline (now creating socket for db before run) and pulled muster

Issue #1380, opened 2025-07-15
Delete sequences from QDB with DROP REFERENCE RELATION stmt


reshke:
$subj

Denchick:
**Current behavior**:
When executing `DROP REFERENCE RELATION rel_name`, the sequence associated with the reference table is not removed.
Currently, it can only be deleted manually by `DROP SEQUENCE sequence_name`.

**Expected behavior**:
DROP REFERENCE RELATION should also remove any sequences owned by the dropped reference table, so they do not remain orphaned in the schema.

**Why this is important:**

- Prevents accumulation of orphaned sequences
- Keeps schema clean
- Matches expected drop behavior for dependent objects

Denchick:
@xelavopelk thank you!

PR #1587, opened 2025-10-05
fix sequence issues

xelavopelk:
- remove seq value when seq deleted in memdb
- fix deleting sequences from QDB with DROP REFERENCE RELATION issue #1380


reshke (approved):

Denchick:
LGTM! Thank you for your contrinution!

Denchick (approved):

PR #1595, opened 2025-10-10
thread safe attach relation to distribuiton

xelavopelk:
For etcd, i have made the relation attach operation to distribution  more thread-safe. The value of a distributed key is now updated using the optimistic lock model, which will be useful when issue #1579 is fixed.

### DIFF
--- a/qdb/models.go
+++ b/qdb/models.go
@@ -116,7 +116,12 @@ type Distribution struct {
 	ColTypes []string `json:"col_types,omitempty"`
 
 	Relations map[string]*DistributedRelation `json:"relations"`
-	Version   int64                           `json:"-"` //if 0 then not exists
+	// The `version` is obtained from the QDB key version.
+	// This field is needed to prevent corruption of the value
+	// when it is updated in concurrent mode by the router.
+	// There is no need to try to deserialize the value or serialize it.
+	// if Version 0 then it not exists in qdb.
+	Version int64 `json:"-"`
 }
 
 type ReferenceRelation struct {


qdb/models.go
```diff
@@ -116,6 +116,7 @@ type Distribution struct {
 	ColTypes []string `json:"col_types,omitempty"`
 
 	Relations map[string]*DistributedRelation `json:"relations"`
+	Version   int64                           `json:"-"` //if 0 then not exists
```
Denchick: I think you might find it useful to include information in the comments about why some entities have the version attribute but others don't

xelavopelk: +

Denchick:
I might be a bit slow here — could you please explain what’s happening in this piece of code?

What confuses me:

- What happens if the key doesn’t have a Version at all (for example, when upgrading an existing installation)?
- What happens if the version doesn’t match (concurrent update scenario)?
- Why do we need WithCountOnly there?

xelavopelk:
> I might be a bit slow here — could you please explain what’s happening in this piece of code?
> 
> What confuses me:
> 
>     * What happens if the key doesn’t have a Version at all (for example, when upgrading an existing installation)?
> 
>     * What happens if the version doesn’t match (concurrent update scenario)?
> 
>     * Why do we need WithCountOnly there?

"**What happens if the key doesn’t have a Version at all (for example, when upgrading an existing installation)?"**

every key in etcd have version which starts from 1. It's  service field " Creating a key increments the version of that key, starting at 1 if the key does not exist at the current revision. Deleting a key generates a key tombstone, concluding the key’s current generation by resetting its version to 0. Each modification of a key increments its version; so, versions are monotonically increasing within a key’s generation." https://etcd.io/docs/v3.5/learning/data_model/


**"What happens if the version doesn’t match (concurrent update scenario)?"**
https://github.com/etcd-io/etcd/blob/client/v3.6.5/client/v3/txn.go#L36
if version does not match we go to branch
https://etcd.io/docs/v3.6/dev-guide/api_reference_v3/
if !resp.Succeeded {
"succeeded is set to true if the compare evaluated to true or false otherwise."

**"Why do we need WithCountOnly there?"**
i need key only to print key version in error message 
"currentVersion := rng.Kvs[0].Version
return fmt.Errorf("distribution was changed '%s' current version is=%d", distribution.ID, currentVersion)"

PR #1596, opened 2025-10-11
partially fixed issue 1579

xelavopelk:
issue #1579 fixed for reference relations
TODO: 
- fix for sharded relations


pkg/coord/coord.go
```diff
@@ -134,7 +147,7 @@ func (lc *Coordinator) CreateReferenceRelation(ctx context.Context, r *rrelation
 	}
 
 	if _, err := lc.qdb.GetReferenceRelation(ctx, relName); err == nil {
-		return fmt.Errorf("reference relation %+v already exists", r.TableName)
+		return errors.New(MessageReferenceRelationExists(r.TableName))
```
Denchick: why? it somehow happened in our repository that fmt.Errorf is used more often errors package.

router/rmeta/coord_remote_master.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, sh := range shs {
+		shardIds = append(shardIds, sh.ID)
+	}
+	newReferenceRelation := &rrelation.ReferenceRelation{
+		TableName:     clauseNode.RelationName,
+		SchemaVersion: 1,
+		ShardIds:      shardIds,
+	}
+	masterCoordinatorConn, err := getMasterCoordinatorConn(ctx, localMngr)
+	if err != nil {
```
Denchick, line 58: This is not a golang-way error handling. Just use [errors.Is](https://pkg.go.dev/errors#example-Is)

pkg/coord/coord.go
xelavopelk: +

router/rmeta/coord_remote_master.go
xelavopelk: +

PR #1629, opened 2025-10-22
update SHOW command

Denchick:


test/feature/features/coordinator.feature
```diff
@@ -774,7 +774,7 @@ Feature: Coordinator test
     Then command return code should be "0"
     When I run SQL on host "coordinator"
     """
-    SHOW task_group
+    SHOW task_groups
```
EinKrebs, line 777: The change is misleading, as there can be but one task group

PR #1631, opened 2025-10-22
Generate move task group on the go without saving all to etcd

EinKrebs:


xelavopelk:
Maybe it would be worth adding something to the documentation.

pkg/models/tasks/tasks.go
```diff
@@ -368,26 +366,22 @@ func MoveTaskToDb(task *MoveTask) *qdb.MoveTask {
...
-		ShardToId: group.ShardToId,
-		KrIdFrom:  group.KrIdFrom,
-		KrIdTo:    group.KrIdTo,
-	}, nil
+		Type:        SplitType(group.Type),
+		ShardToId:   group.ShardToId,
+		KrIdFrom:    group.KrIdFrom,
+		KrIdTo:      group.KrIdTo,
+		BoundRel:    group.BoundRel,
+		Coeff:       group.Coeff,
```
Denchick, line 379: interesting, who is mr coeff?

Denchick (approved):

Denchick:
> Maybe it would be worth adding something to the documentation.

hahaha, we have like a small emergency situation here, I plan to return here soon to write the missing documentation.

PR #1632, opened 2025-10-23
fix: autocreate distributed tables 

xelavopelk:
autocreate distributed tables added
issue #1579


router/rmeta/coord_remote_master.go
```diff
@@ -79,3 +77,66 @@ func CreateReferenceRelation(ctx context.Context, localMngr meta.EntityMgr, clau
...
+	if len(distribution.ColTypes) != 1 {
+		return fmt.Errorf("automatic attach is supported only for distribution with one column key")
+	}
+	if distribution.ColTypes[0] != qdb.ColumnTypeVarchar &&
+		distribution.ColTypes[0] != qdb.ColumnTypeInteger &&
+		distribution.ColTypes[0] != qdb.ColumnTypeUUID {
+		return fmt.Errorf("automatic attach isn't supported for column key %s", distribution.ColTypes[0])
+	}
+	err = mngr.AlterDistributionAttach(ctx, distributionId, rels)
+	return err
```
EinKrebs, line 116: Can you explain, why can't we automatically attach any other column type?

router/rmeta/coord_remote_master.go
```diff
@@ -79,3 +77,66 @@ func CreateReferenceRelation(ctx context.Context, localMngr meta.EntityMgr, clau
...
+			ReplicatedRelation: false,
+			DistributionKey: []distributions.DistributionKeyEntry{
+				{
+					Column: distributionKey,
+					/* support hash function here */
+				},
+			},
+		},
+	}
+	masterCoordinatorConn, err := getMasterCoordinatorConn(ctx, localMngr)
```
EinKrebs: I think it's better to use coordinatorAdapter here, as it's specifically designed for doing proxy coordinator operation on routers

EinKrebs (changes requested):
See comments above

router/rmeta/coord_remote_master.go
xelavopelk: adapter is used:

masterCoordinator := coord.NewAdapter(masterCoordinatorConn)

router/rmeta/coord_remote_master.go
xelavopelk: we need got hashfunction from somewhere
PS
```
					Column: distributionKey,
					/* support hash function here */
```

router/rmeta/coord_remote_master.go
xelavopelk: this struct is used below:
`coord.NewAdapter(masterCoordinatorConn)`
On the other hand, you're right. This place looks ugly. I'll try beautify this function.

xelavopelk: beautified

xelavopelk:
> See comments above

answered both comments

router/rmeta/coord_remote_master.go
```diff
@@ -2,33 +2,49 @@ package rmeta
...
+// - ctx: (context.Context): context
+// - localCoordinator (meta.EntityMgr): Current (local) coordinator
+//
+// Returns:
+// - masterMngr (meta.EntityMgr) master coordinator
+// - close (func() error) function for closing master coordinator connection.
+// - error: An error when fails.
+func getMasterCoordinator(ctx context.Context,
+	localCoordinator meta.EntityMgr) (masterMngr meta.EntityMgr, close func() error, err error) {
+	noNeedClose := func() error { return nil }
```
Denchick: No need to close

docs/routing/hints.mdx
```diff
@@ -182,7 +182,7 @@ NOTICE: send query to shard(s) : sh1,sh2,sh3,sh4
 ### __spqr__distribution_key
 
 The `__spqr__distribution_key` parameter is used to automatically add a new table to the SPQR metadata without having to go into the spqr console. This parameter can only be used with the `__spqr__auto_distribution` hint.
-
+Distributions with a single key column are now supported. The following key types are supported for automatic creation of distributed tables: integer, UUID, and varchar. Hashed distributions are currently not supported now for autocreate tables.
```
Denchick:
```
<Warning>Distributions with a single key column are only supported.<Warning>

<Warning>The following key types are supported for the automatic creation of distributed tables: `integer`, `uuid`, and `varchar`.</Warning>

<Warning>Hash distributions are currently not supported for auto-creating tables.</Warning>
```

Denchick (changes requested):
Looks great overall! Just a few minor comments and suggestions below.

router/rmeta/coord_remote_master.go
xelavopelk: done

docs/routing/hints.mdx
xelavopelk: done

reshke:
Please rebase, you will have to deal with coord_remote_master file removal.
I did this mainly for import hazards: rmeta should not depend on adapter-grpc-coord etc...

reshke:
Ideally, most of the code should move to console like in https://github.com/pg-sharding/spqr/commit/2f09eef77ce6702bf65a6353865d88948e2c9f51
Also, maybe we can avoid logic duplication with ProcMetadata somehow

xelavopelk:
> Please rebase, you will have to deal with coord_remote_master file removal. I did this mainly for import hazards: rmeta should not depend on adapter-grpc-coord etc...

It may be better to use the config key `use_coordinator_init` to determine whether a master coordinator is needed instead of
```
coordAddr, err := localCoordinator.GetCoordinator(ctx)
...
if coordAddr == "" {
    return localCoordinator, noNeedToClose, nil
  }
```

xelavopelk:
v2: https://github.com/pg-sharding/spqr/pull/1848

PR #1665, opened 2025-10-30
refactor unlocks

xelavopelk:
0) Now it is threadsafe. Refactored retry in unlock: the aim was to remove non serializable business logic from the QDB-layer. See the comment in "defer unlockkeyrange" for why the approach differed from "Lock".
1) Apply nowaitlock to the "split key range" command.

### DIFF
--- a/pkg/coord/coord.go
+++ b/pkg/coord/coord.go
@@ -3,7 +3,6 @@ package coord
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"github.com/google/uuid"
 	"github.com/pg-sharding/spqr/pkg/config"
@@ -749,7 +748,7 @@ func (qc *Coordinator) LockKeyRange(ctx context.Context, keyRangeID string) (*kr
 // Returns:
 // - error: an error if the unlock operation encounters any issues.
 func (qc *Coordinator) UnlockKeyRange(ctx context.Context, keyRangeID string) error {
-	return retry.Do(ctx, retry.NewFibonacci(qdb.LockRetryStep*time.Millisecond),
+	return retry.Do(ctx, retry.NewFibonacci(qdb.LockRetryStep),
 		func(ctx context.Context) error {
 			return qc.qdb.UnlockKeyRange(ctx, keyRangeID)
 		})
--- a/qdb/qdb.go
+++ b/qdb/qdb.go
@@ -3,14 +3,13 @@ package qdb
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/pg-sharding/spqr/pkg/config"
 	"github.com/pg-sharding/spqr/router/rfqn"
 )
 
 const LockRetryStep = 500 * time.Millisecond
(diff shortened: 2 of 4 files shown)


test/etcdqdb_integration/etcdqdb_test.go
```diff
@@ -112,3 +113,53 @@ func TestQdb(t *testing.T) {
...
+			is.NoError(err)
+			_, err = db.NoWaitLockKeyRange(ctx, keyRange1.KeyRangeID)
+			is.NoError(err)
+			err = db.UnlockKeyRange(ctx, keyRange1.KeyRangeID)
+			is.NoError(err)
+			_, err = db.CheckLockedKeyRange(ctx, keyRange1.KeyRangeID)
+			expectedErr := spqrerror.Newf(spqrerror.SPQR_KEYRANGE_ERROR, "key range %v not locked", keyRange1.KeyRangeID)
+			is.Equal(expectedErr, err)
+		})
+		t.Run("unlock on not locked kr", func(t *testing.T) {
```
Denchick: Maybe it's better to make the unlock key range identipotent?

qdb/qdb.go
```diff
@@ -8,6 +8,10 @@ import (
 	"github.com/pg-sharding/spqr/router/rfqn"
 )
 
+const (
```
Denchick:
```suggestion
const LockRetryStep = 500 * time.Millisecond
```

pkg/coord/coord.go
```diff
@@ -892,14 +899,16 @@ func (qc *Coordinator) Split(ctx context.Context, req *kr.SplitKeyRange) error {
...
 	}
 
-	krOldDB, err := qc.qdb.LockKeyRange(ctx, req.SourceID)
+	krOldDB, err := qc.qdb.NoWaitLockKeyRange(ctx, req.SourceID)
 	if err != nil {
 		return err
 	}
 
 	if !req.SplitLeft {
+		// TODO: after convert split command into etcd transaction we no need in embracing "lock" "unlock".
```
Denchick, line 907: I don't understand, why did you live these TODOs? I just don't understand what is action item for them? What should we do to fix them i mean

xelavopelk: > I don't understand, why did you live these TODOs? I just don't understand what is action item for them? What should we do to fix them i mean

when "split" operation will be converted into etcd transaction we'll no need exec embrace this this operation "lock"-"unlock"commands. We just need check lock at the start operation in etcd transaction.

xelavopelk: it's mean we'll no nead any  retry when operation will be atomic and transactional.

test/etcdqdb_integration/etcdqdb_test.go
xelavopelk: done

qdb/qdb.go
xelavopelk: done

Denchick (approved):
LGTM!

PR #1724, opened 2025-11-09
add/drop shard are implemented in adapter

xelavopelk:
issue #1725

### DIFF
--- a/pkg/coord/adapter.go
+++ b/pkg/coord/adapter.go
@@ -697,7 +697,7 @@ func (a *Adapter) AddDataShard(ctx context.Context, shard *topology.DataShard) e
 // - error: An error if the data shard drop fails, otherwise nil.
 func (a *Adapter) DropShard(ctx context.Context, shardId string) error {
 	client := proto.NewShardServiceClient(a.conn)
-	_, err := client.DropDataShard(ctx, &proto.ShardRequest{Id: shardId})
+	_, err := client.DropShard(ctx, &proto.DropShardRequest{Id: shardId})
 	return err
 }
 
--- a/pkg/protos/shard.pb.go
+++ b/pkg/protos/shard.pb.go
@@ -413,15 +413,14 @@ const file_protos_shard_proto_rawDesc = "" +
 	"\x14AddWorldShardRequest\x12!\n" +
 	"\x05shard\x18\x01 \x01(\v2\v.spqr.ShardR\x05shard\"\"\n" +
 	"\x10DropShardRequest\x12\x0e\n" +
-	"\x02id\x18\x01 \x01(\tR\x02id2\x87\x03\n" +
+	"\x02id\x18\x01 \x01(\tR\x02id2\xc8\x02\n" +
 	"\fShardService\x12=\n" +
 	"\n" +
 	"ListShards\x12\x16.google.protobuf.Empty\x1a\x15.spqr.ListShardsReply\"\x00\x12?\n" +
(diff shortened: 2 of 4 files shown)


coordinator/pkg/clustered_coord.go
```diff
@@ -2091,9 +2091,44 @@ func (qc *ClusteredCoordinator) ProcClient(ctx context.Context, nconn net.Conn,
...
+		}
+
+		spqrlog.Zero.Debug().
+			Interface("response", resp).
+			Msg("add shard response")
+		return nil
+	})
+}
+
+func (qc *ClusteredCoordinator) DropShard(ctx context.Context, shardId string) error {
```
Denchick, line 2268: wait, wat? There was no DropShard before???

test/feature/features/proxy_console.feature
```diff
@@ -302,4 +302,38 @@ Feature: Proxy console
...
             "Temporary key range ID":   "temp_id"
         }]
-        """
\ No newline at end of file
+        """
+    
+    Scenario: add and drop shard on router-admin
+        When I run SQL on host "router-admin"
+        """
+        CREATE SHARD sh5 WITH HOSTS klepov:6432;
```
Denchick, line 316: I guess `klepov:6432` doesn't exist.
Are we sure we want to add a shard that is unreachable?
For example, at the start of the router, we check that each shard is reachable. If it is not, the router will not start. (What? Yes!)

diPhantxm:
Hi!
Does router list shards from coordinator or from local config? It's important because when we add new shard via coordinator, coordinator will ask routers to add this new shard. But after restart router seems to read shards from config and not from coordinator, according to this https://github.com/pg-sharding/spqr/blob/master/router/instance/instance.go#L143
This will result in some misconfiguration because router will not know about all shards that coordinator knows. I was working on shards management via coordinator in this [pr](https://github.com/pg-sharding/spqr/pull/1518) but wasn't managed to finish the work

xelavopelk:
> Hi! Does router list shards from coordinator or from local config? It's important because when we add new shard via coordinator, coordinator will ask routers to add this new shard. But after restart router seems to read shards from config and not from coordinator, according to this https://github.com/pg-sharding/spqr/blob/master/router/instance/instance.go#L143 This will result in some misconfiguration because router will not know about all shards that coordinator knows. I was working on shards management via coordinator in this [pr](https://github.com/pg-sharding/spqr/pull/1518) but wasn't managed to finish the work

So command "regress=> CREATE SHARD sh5 WITH HOSTS big_ass:6432;" on coordinator works incorrect too?

xelavopelk:
> Hi! Does router list shards from coordinator or from local config? It's important because when we add new shard via coordinator, coordinator will ask routers to add this new shard. But after restart router seems to read shards from config and not from coordinator, according to this https://github.com/pg-sharding/spqr/blob/master/router/instance/instance.go#L143 This will result in some misconfiguration because router will not know about all shards that coordinator knows. I was working on shards management via coordinator in this [pr](https://github.com/pg-sharding/spqr/pull/1518) but wasn't managed to finish the work

ok when i add shard on coordinator(on master!!!!!) it's router don't show in.
it's master:

coordinator:
image
router:
image

test/feature/features/proxy_console.feature
Denchick: https://github.com/pg-sharding/spqr/issues/1764

xelavopelk:
cluster mode in add/drop shard not works. blocker: https://github.com/pg-sharding/spqr/pull/1518

PR #1739, opened 2025-11-11
Fix regress tests

EinKrebs:


test/regress/run_tests.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
```
xelavopelk: may be need change in 'run_tests_coord.sh'

EinKrebs: `set -e` caused problems with the script, so addressed the issue differently

PR #1744, opened 2025-11-12
fix reload.feature test

xelavopelk:
fixed "reload.feature". I don't understand too why it worked before.


Denchick:
@xelavopelk thank you for your contribution!

PR #1750, opened 2025-11-12
Add tests for reversed equality and IN-list routing in planByQualExpr

cypny:
- Handle Const = ColumnRef in planByQualExpr
- Enabled and fix reversed equality case ('const' = id)
- Enabled and fix IN-list scatter test


Denchick:
@cypny thank you for your contribution!

PR #1752, opened 2025-11-13
Refactor `rewrite query` support for reference relation modification

reshke:


pkg/plan/plan.go
```diff
@@ -12,7 +15,7 @@ type Plan interface {
 	Stmt() lyx.Node
 	SetStmt(lyx.Node)
 	ExecutionTargets() []kr.ShardKey
-	GetQuery(sh string) string
+	GetGangMemberMsg(sh kr.ShardKey) pgproto3.FrontendMessage
```
Denchick: here we go 
could you please add somewhere who the gang is?

PR #1777, opened 2025-11-23
grpc error style fixed

xelavopelk:
0) small fix clustered regress run
1)  Introduced an grpc-error returning in the google-style and handled it on the adapter. As result removed redundant "rpc error: code = Unknown desc". Removed allerts "rpc error" on "console" regress test for clustered installation (see file
[cluster_regress_result_20251123.txt](https://github.com/user-attachments/files/23696335/cluster_regress_result_20251123.txt)
).
PS If ok need fix all grpc endpoints.


test/regress/Dockerfile_coord
```diff
@@ -22,7 +22,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 ENV PATH=$PATH:"/usr/lib/postgresql/${POSTGRES_VERSION}/lib/pgxs/src/test/regress:/home/build-user/pgbin/lib/postgresql/pgxs/src/test/regress"
 
 COPY ./ /regress
-RUN chmod a+x /regress/run_tests.sh
```
reshke, line 25: why

Denchick: I guess because: "PS If ok need fix all grpc endpoints."

Denchick:
> Introduced an grpc-error returning in the google-style and handled it on the adapter. As result removed redundant "rpc error: code = Unknown desc". Removed allerts "rpc error" on "console" regress test for clustered installation (see file
[cluster_regress_result_20251123.txt](https://github.com/user-attachments/files/23696335/cluster_regress_result_20251123.txt)
).

@xelavopelk Maybe do you have some examples? Like before/after?

xelavopelk:
> > Introduced an grpc-error returning in the google-style and handled it on the adapter. As result removed redundant "rpc error: code = Unknown desc". Removed allerts "rpc error" on "console" regress test for clustered installation (see file
> > [cluster_regress_result_20251123.txt](https://github.com/user-attachments/files/23696335/cluster_regress_result_20251123.txt)
> > ).
> 
> @xelavopelk Maybe do you have some examples? Like before/after?

regress_tests_coord   | -ERROR:  distribution "ds23223" not found
regress_tests_coord   | +ERROR:  rpc error: code = Unknown desc = distribution "ds23223" not found
...
regress_tests_coord   | -ERROR:  failed to split because bound is out of key range
regress_tests_coord   | +ERROR:  rpc error: code = Unknown desc = failed to split because bound is out of key range

test/regress/Dockerfile_coord
xelavopelk: because file name Dockerfile_coord - test with coordinator installation

xelavopelk: For this time regress test with clustered installation is unstable. I propose keep it the distanse on current regress test.

Issue #1716, opened 2025-11-07
Add built-in Prometheus metrics endpoint


Denchick:
Currently, `spqr-router` and `spqr-coordinator` does not expose any metrics over HTTP.
It would be extremely helpful to have a built-in `/metrics` endpoint exporting router/coordinator/both statistics in Prometheus-compatible format.

Although many useful runtime stats are already available via the [`SHOW` command](https://docs.pg-sharding.tech/sharding/console/sql_commands#show), they are only accessible via SQL and cannot be easily integrated with standard monitoring systems.

I generated some examples:

```
var (
	activeClientConnectionsMetric = promauto.NewGauge(prometheus.GaugeOpts{
		Name: "spqr_active_client_connections",
		Help: "The number of active client connections (from SHOW clients).",
	})

	activeBackendConnectionsMetric = promauto.NewGauge(prometheus.GaugeOpts{
		Name: "spqr_active_backend_connections",
		Help: "The number of active backend connections (from SHOW backend_connections).",
	})

	routerUptimeSecondsMetric = promauto.NewGauge(prometheus.GaugeOpts{
		Name: "spqr_router_uptime_seconds",
		Help: "Router uptime in seconds (derived from SHOW status).",
	})

	poolSizeMetric = promauto.NewGaugeVec(prometheus.GaugeOpts{
		Name: "spqr_pool_size",
		Help: "The number of connections in each pool (from SHOW pools).",
	}, []string{"pool_name"})

	queryCountMetric = promauto.NewCounter(prometheus.CounterOpts{
		Name: "spqr_queries_total",
		Help: "Total number of SQL queries processed by the router (derived from SHOW time_quantiles or logs).",
	})

	queryLatencyQuantilesMetric = promauto.NewHistogram(prometheus.HistogramOpts{
		Name:    "spqr_query_latency_seconds",
		Help:    "Query latency distribution (based on SHOW time_quantiles).",
		Buckets: prometheus.DefBuckets,
	})

	moveTasksInProgressMetric = promauto.NewGauge(prometheus.GaugeOpts{
		Name: "spqr_move_tasks_in_progress",
		Help: "Number of ongoing move tasks (from SHOW move_task).",
	})
)

```

Denchick:
I thought a lot about advantages and disadvantages of this issue and came up to the conclusion we shouldn't implement it. Why?

1. Prometheus is not PostgreSQL-way
[...]

PR #1827, opened 2025-12-07
Prometheus metrics

Denchick:
https://github.com/pg-sharding/spqr/issues/1716


Denchick:
https://github.com/pg-sharding/spqr/issues/1716

PR #1835, opened 2025-12-11
refactoring CreateDistribution for 2 phase execution

xelavopelk:
- Field Version in struct qdb.Distribution is removed. After implementation atomic operations we not need for it.
- The router and coordinator exchange grpc-services have been completed. When testing CreateDistribution, it was found that insufficient code had been transferred from the PoC).
- implemented 2 phase execution flow (see issue #1737) for CreateDistribution command


coordinator/pkg/clustered_coord.go
```diff
@@ -2268,17 +2275,12 @@ func (qc *ClusteredCoordinator) DropReferenceRelation(ctx context.Context,
...
-// CreateDistribution creates distribution in QDB
-// TODO: unit tests
-func (qc *ClusteredCoordinator) CreateDistribution(ctx context.Context, ds *distributions.Distribution) error {
-	if err := qc.Coordinator.CreateDistribution(ctx, ds); err != nil {
-		return err
-	}
-
-	return qc.traverseRouters(ctx, func(cc *grpc.ClientConn) error {
+/*
+func gossipCreateDistribution(gossip *proto.CreateDistributionGossip) func(cc *grpc.ClientConn) error {
```
Denchick: Why not just remove it?

coordinator/pkg/clustered_coord.go
```diff
@@ -2288,7 +2290,25 @@ func (qc *ClusteredCoordinator) CreateDistribution(ctx context.Context, ds *dist
...
+}*/
+
+// CreateDistribution creates distribution in QDB
+// TODO: unit tests
+func (qc *ClusteredCoordinator) CreateDistribution(ctx context.Context, ds *distributions.Distribution) (*mtran.MetaTransactionChunk, error) {
+	if transactionChunk, err := qc.Coordinator.CreateDistribution(ctx, ds); err != nil {
+		return nil, err
+	} else {
+		if len(transactionChunk.GossipRequests) > 0 {
+			return nil, fmt.Errorf("clustered coord have got gossip requests from local coord (case 0)")
```
Denchick:
```suggestion
			return nil, fmt.Errorf("local coordinator unexpectedly returned gossip requests; expected empty gossip requests from local coordinator")
```

coordinator/provider/distributions.go
```diff
@@ -24,17 +25,28 @@ func NewDistributionServer(impl coordinator.Coordinator) *DistributionsServer {
...
 	for _, ds := range req.Distributions {
 		mds, err := distributions.DistributionFromProto(ds)
 		if err != nil {
 			return nil, err
 		}
-		if err := d.impl.CreateDistribution(ctx, mds); err != nil {
+		if tranChunk, err := d.impl.CreateDistribution(ctx, mds); err != nil {
 			return nil, err
+		} else {
+			if len(tranChunk.QdbStatements) == 0 {
```
Denchick, line 39: May tranChunk be nil?

coordinator/provider/transactions.go
```diff
@@ -0,0 +1,66 @@
...
+func (mts *MetaTransactionServer) ExecNoTran(ctx context.Context, request *proto.ExecNoTranRequest) (*emptypb.Empty, error) {
+	if stmts, err := toQdbStatementList(request.CmdList); err != nil {
+		return nil, err
+	} else {
+		if tranChunk, err := mtran.NewMetaTransactionChunk(request.MetaCmdList, stmts); err != nil {
+			return nil, err
+		} else {
+			return nil, mts.impl.ExecNoTran(ctx, tranChunk)
+		}
+	}
```
Denchick:
```suggestion
    stmts, err := toQdbStatementList(request.CmdList)
    if err != nil {
        return nil, err
    }
    tranChunk, err := mtran.NewMetaTransactionChunk(request.MetaCmdList, stmts)
    if err != nil {
        return nil, err
    }
    return nil, mts.impl.ExecNoTran(ctx, tranChunk)
```

coordinator/provider/distributions.go
xelavopelk: no. it cant't be empty too

xelavopelk: it can't be null. every change command must generate not null chunk if error!=nil. 

Moreover, an empty chunk is a logical error. we are checking this.

coordinator/pkg/clustered_coord.go
xelavopelk: done (make generate)

Denchick:
In this PR, we are waiting for approval from @reshke. Our plan is to first check https://github.com/pg-sharding/spqr/pull/1847 in production, then merge this PR.

PR #1837, opened 2025-12-11
Add shard-match to DefaultRouteBehaviour.

reshke:
Implement something that should be here for a long time ago.
Somehow no one asked for it until now.


reshke:
Implement something that should be here for a long time ago.
Somehow no one asked for it until now.

Issue #1817, opened 2025-12-04
Add __spqr__engine_v2 to the router config


Denchick:
**Problem**: there is __spqr__engine_v2 routing hint but there is no way to set it by default in router.yaml.
**Solution**: add such setting, for example: `query_routing.engine_default: v2`

Denchick:
Sorry, the issue seems to be invalid.

https://docs.pg-sharding.tech/routing/hints#spqr-engine-v2

PR #1849, opened 2025-12-16
Feature/add   spqr  engine v2 to the router config

KKittyCatik:
Problem: there is __spqr__engine_v2 routing hint but there is no way to set it by default in router.yaml.
Solution: add such setting, for example: query_routing.engine_default: v2


Denchick:
Thank you for your time. But the original issue seems to be invalid https://github.com/pg-sharding/spqr/issues/1817.

PR #1870, opened 2025-12-30
FIx concurrent map writes in clustered coordiantor

EinKrebs:


coordinator/pkg/clustered_coord.go
```diff
@@ -250,8 +250,9 @@ type ClusteredCoordinator struct {
 	cache        *cache.SchemaCache
 	acquiredLock bool
 
-	bounds map[string][][][]byte
-	index  map[string]int
+	boundMutex sync.RWMutex
```
reshke: sync map?

PR #1874, opened 2026-01-01
key range validation is refactored

xelavopelk:
Key range checks moved from qdb layer to meta layer. Preparing to convert CreateKeyRange to 2 phase execution.
issue #1737

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -401,7 +401,6 @@ phandler
 pitr
 pkgclient
 pmgr
-Pmgr
 podman
 poolers
 poolmgr
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -12,7 +12,7 @@ import (
 	"github.com/pg-sharding/spqr/router/statistics"
 )
 
-type Pmgr interface {
+type ParamManager interface {
 	SetParam(string, string)
 	ResetParam(string)
 	ResetAll()
@@ -28,7 +28,7 @@ type Pmgr interface {
 }
 
 type Client interface {
-	Pmgr
+	ParamManager
 	session.SessionParamsHolder
 	statistics.StatHolder
 


pkg/meta/validators/kr_validators.go
```diff
@@ -0,0 +1,111 @@
...
+		return spqrerror.Newf(spqrerror.SPQR_KEYRANGE_ERROR, "key range %v equals key range %v in QDB", keyRange.ID, nearestKr.ID)
+	}
+	if nearestKr != nil && nearestKr.ShardID != keyRange.ShardID {
+		return spqrerror.Newf(spqrerror.SPQR_KEYRANGE_ERROR, "key range %v intersects with key range %v in QDB", keyRange.ID, nearestKr.ID)
+	}
+
+	return nil
+}
+
+// ValidateKeyRangeForCreate validates key range before modifying
```
Denchick:
```suggestion
// ValidateKeyRangeForModify validates key range before modifying
```

coordinator/pkg/clustered_coord.go
```diff
@@ -656,7 +656,11 @@ func (qc *ClusteredCoordinator) CreateKeyRange(ctx context.Context, keyRange *kr
 
 // TODO : unit tests
 func (qc *ClusteredCoordinator) MoveKeyRange(ctx context.Context, keyRange *kr.KeyRange) error {
-	return ops.ModifyKeyRangeWithChecks(ctx, qc.db, keyRange)
+	// TODO: move check to meta layer
```
Denchick, line 659: Should this TODO be removed now that the check IS in the meta layer (validators package)?

```
// TODO: move check to meta layer
```

Same comment applies to the same TODOs in this file

pkg/meta/validators/kr_validators_test.go
```diff
@@ -0,0 +1,188 @@
...
+		return nil, err
+	}
+	if err = memqdb.AddShard(ctx, mockShard2); err != nil {
+		return nil, err
+	}
+	return memqdb, nil
+}
+
+func TestValidateKeyRangeForCreate_happyPath(t *testing.T) {
+	assert := assert.New(t)
```
Denchick: Nitpick: the test file uses `assert := assert.New(t)` style while most other tests in the codebase use `assert.NoError(t, err)` directly. Not blocking, just noting for consistency.

coordinator/pkg/clustered_coord.go
xelavopelk: I changed the check so that it does not depend on QDB, but on EntityMngr. This should be fixed when the MoveKeyRange function switches to two-phase mode.

pkg/meta/validators/kr_validators_test.go
xelavopelk: ok

Issue #1810, opened 2025-12-02
Add HINT to MOVE KEY RANGE command


Denchick:
Problem: it is not clear for newcomers that MOVE KEY RANGE actually move metadata only, it is not move data itself. We note it in the documention but still.
Expected behaviour:

```
MOVE KEY RANGE krid3 to sh2
HINT: If you don’t fully understand how SPQR works, maybe you need REDISTRIBUTE KEY RANGE
```

Denchick:
Fixed in https://github.com/pg-sharding/spqr/pull/1884

PR #1884, opened 2026-01-04
Feat: MOVE KEY RANGE command  hint (#1810)

kla1mn:
PR for #1810 issue


kla1mn:
> What do you think?

Yeah, I agree, it's better to emphasize the behavioral difference. I'll change the hint text and test for it, respectively.

Denchick (approved):
LGTM.

PR #1899, opened 2026-01-09
Derive CommandComplete message for output slice in executor.

reshke:
For now, we only have single-slice queries, so derive is trivial reuse.

### DIFF
--- a/router/relay/relay.go
+++ b/router/relay/relay.go
@@ -219,13 +219,14 @@ func (rst *RelayStateImpl) Reset() error {
 		return err
 	}
 
+	rst.QueryExecutor().ActiveShardsReset()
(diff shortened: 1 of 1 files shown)


rkhapov (approved):

reshke:
commit msg:


For now, we only have single-slice queries, so derive is trivial reuse. Only exception case in emptyMessage/emptyMessageResponse, which has no CommandComplete message in interaction loop.
Also we still have some troubles in multi-statement processing, which starting from this commit only gets worse.
It does not however induce any new regressions, only complicates status-quo. 
The main deal-breaker for multi-statement support is existence of ParseState, but now we also need to support 
multiple CommandComplete message if executor, or do DeriveCC call more frequently.

PR #1919, opened 2026-01-14
Introduce __spqr__ctid function.

reshke:
This commit introduces __spqr__ctid function which returns whole relation contents with additional prefixed column `__spqr__ctid` which describes where data comes from. This is first step to full support for virtual execution, which will allow to get per-shard aggregates to work.

Per long-standing suggestion from Alexey Rybak

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -87,6 +87,7 @@ cptr
 createsequence
 Cta
 CTEs
+ctid
 ctl
 ctsa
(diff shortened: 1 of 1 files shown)


reshke:
This commit introduces __spqr__ctid function which returns whole relation contents with additional prefixed column `__spqr__ctid` which describes where data comes from. This is first step to full support for virtual execution, which will allow to get per-shard aggregates to work.

Per long-standing suggestion from Alexey Rybak

PR #1928, opened 2026-01-18
part 2 of "CreateKeyRange refactored for meta transaction

xelavopelk:
0) part changes tran_entity_manager for create range
1) fix for interface bug in transactions.go

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/jcmturner/gokrb5/v8 v8.4.4
 	github.com/lib/pq v1.10.9
 	github.com/libp2p/go-reuseport v0.4.0
-	github.com/pg-sharding/lyx v0.0.0-20260116115458-8cedfa2a7b7f
+	github.com/pg-sharding/lyx v0.0.0-20260119082018-b5c60c949352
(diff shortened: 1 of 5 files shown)


pkg/meta/tran_entity_manager.go
```diff
@@ -254,9 +273,30 @@ func (t *TranEntityManager) ListKeyRangeLocks(ctx context.Context) ([]string, er
...
+	if err != nil {
+		return nil, err
+	}
+	result := make([]*kr.KeyRange, 0, len(list)+len(t.keyRanges.Items()))
+	for _, keyRange := range t.keyRanges.Items() {
+		if keyRange.Distribution == distribution {
+			result = append(result, keyRange)
+		}
+	}
+	for _, keyRange := range list {
```
Denchick, line 289: Consider pre-computing deleted and transaction items to avoid repeated map lookups in the loop

Denchick (approved):

pkg/meta/tran_entity_manager.go
xelavopelk: just O(n)

PR #1970, opened 2026-01-26
More SPQR isolation test

reshke:


EinKrebs:
@reshke pls rebase after #1986

PR #1974, opened 2026-01-27
Simplification network exchange during meta transaction

xelavopelk:
0. QDB statements have been removed from the meta<->coordinator exchange during  meta transaction.
1. The conversion of gossip commands to QDB statements is now used not only in routers, but also in coordinators. This business logic has been moved to a common file called coord_meta_tran.go.
2. Several structures that are no longer needed have been marked for removal  "NEED REMOVE. This structure doesn't make sense right now.". They will be removed next PR.
3. The TranEntityManager has been updated to encapsulate the business logic for transactions on the meta side.
4. We are considering separating the EntityMgr interface at the meta and coordinator levels. At the meta level, we use for write commands either commitTransaction or ExecNoTran exclusively.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -214,6 +214,7 @@ IDR
 IDTo
 idxs
 IExecute
+IKill
 illum
 incididunt
 inds
--- a/coordinator/pkg/clustered_coord.go
+++ b/coordinator/pkg/clustered_coord.go
@@ -48,10 +48,13 @@ import (
 	"github.com/pg-sharding/spqr/router/rfqn"
 	"github.com/pg-sharding/spqr/router/route"
 	spqrparser "github.com/pg-sharding/spqr/yacc/console"
+	"github.com/sethvargo/go-retry"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/keepalive"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/emptypb"
 )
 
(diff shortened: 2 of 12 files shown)


protos/meta_transaction.proto
```diff
@@ -29,7 +29,7 @@ message MetaTransactionGossipRequest {
 service MetaTransactionGossipService {
   rpc ApplyMeta(MetaTransactionGossipRequest) returns (google.protobuf.Empty) {}
 }
-
+// NEED REMOVE. This structure doesn't make sense right now.
```
Denchick, line 32: Like in the next PR or?

xelavopelk: it will be removed next PR.

pkg/coord/coord_meta_tran.go
```diff
@@ -0,0 +1,55 @@
...
+func createDistributionPrepare(ctx context.Context, mngr meta.EntityMgr, gossip *proto.CreateDistributionGossip) ([]qdb.QdbStatement, error) {
+	result := make([]qdb.QdbStatement, 0, len(gossip.GetDistributions()))
+	for _, ds := range gossip.GetDistributions() {
+		mds, err := distributions.DistributionFromProto(ds)
+		if err != nil {
+			return nil, err
+		}
+		if qdbStatements, err := mngr.CreateDistribution(ctx, mds); err != nil {
+			return nil, err
+		} else {
```
Denchick: if-else standard issue :)

here and https://github.com/pg-sharding/spqr/pull/1974/changes#diff-2adb27bfec0f3e3579932d4fb7eb1bff716805ff1cc13fbc1ce1231bc7d9a017R40

pkg/meta/tran_entity_manager.go
```diff
@@ -11,22 +11,65 @@ import (
...
 	meta_transaction "github.com/pg-sharding/spqr/pkg/models/transaction"
+	proto "github.com/pg-sharding/spqr/pkg/protos"
 	"github.com/pg-sharding/spqr/qdb"
 	"github.com/pg-sharding/spqr/router/cache"
 	"github.com/pg-sharding/spqr/router/rfqn"
 )
 
+type TransactionState struct {
+	// prevents double execution
+	isCommitted bool
```
Denchick, line 23: I can't find any usage of this variable

pkg/coord/coord_meta_tran.go
xelavopelk: fixed

pkg/meta/tran_entity_manager.go
xelavopelk: I crookedly chopped off here. Glued it back on.

PR #1975, opened 2026-01-28
Implement REDISTRIBUTE with NOWAIT options.

reshke:
In order to get isolation test in-tree faster, and for overall usability.
Making this a default requires separate effort and discussion.


yacc/console/yx_test.go
```diff
@@ -424,6 +424,17 @@ func TestRedistribute(t *testing.T) {
 			},
 			err: nil,
 		},
+		{
+			query: "REDISTRIBUTE KEY RANGE kr1 TO sh2 BATCH SIZE 500 NOWAIT",
```
Denchick, line 440: how this nowait keyword may i combine with check and apply keywords?

reshke: no way currenly

PR #1992, opened 2026-02-05
Add RPS metric tracking to spqr-console

### LINKED ISSUE #1913: Add RPS metric to spqr-console
**Problem:** Requests Per Second is a fundamental metric for monitoring router performance and capacity planning. Currently, there's no way to observe the throughput of the router in real-time.

**Solution:** Add a metric to track the number of requests per second processed by the router, accessible via spqr-console.

Example:
```
SHOW router_rps;

 current_rps | avg_rps | peak_rps
-------------+---------+----------
        1250 |     980 |     3200
```

Rounakneema:
# Add RPS Metric to spqr-console

## Summary
This PR implements issue #1913 by extending the `SHOW instance` command in spqr-console to display real-time router throughput metrics (Requests Per Second).

## Problem
Requests Per Second (RPS) is a fundamental metric for monitoring router performance and capacity planning. Currently, there is no way to observe the throughput of the SPQR router in real-time.

## Solution
Extended the `SHOW instance` command to include 4 new columns:

| Metric | Description |
|--------|-------------|
| `total_requests` | Lifetime request count (Atomic, always available) |
| `current_rps` | Request rate over the last second (Sliding window) |
| `avg_rps` | Lifetime average: total_requests / seconds_since_start |
| `peak_rps` | Highest current_rps value ever observed |

The expensive sliding window calculation is optional and controlled by configuration.

**Example output:**
```
spqr=> SHOW instance;
 ... | active_tcp_connections | total_requests | current_rps | avg_rps | peak_rps
-----+------------------------+----------------+-------------+---------+----------
 ... |                      1 |           1542 |       12.50 |    9.80 |    32.00
(1 row)
```

### Core Components

1. **RPS Tracking Module** (`router/statistics/rps.go`)
   - **Atomic Counting**: `total_requests` uses lock-free `atomic.AddInt64`.
   - **Sliding Window**: Calculates smooth RPS (1s window, 10 buckets). Controlled by config to avoid mutex contention on high load.
   - **Thread-safe**: Proper concurrent access handling.

2. **Configuration** (`pkg/config/router.go`)
   - Added `router_rps_aggregation` boolean setting (default `false`) to enable/disable the sliding window statistics.

3. **Console Integration** (`pkg/engine/virtual_relation.go`)
   - Updated `InstanceVirtualRelationScan` to fetch and display RPS metrics.

4. **Request Hook** (`router/frontend/frontend.go`)
   - Integrated `statistics.OnRequest()` into the request handling path.

## Files Changed

- `router/statistics/rps.go` (new) - RPS tracking implementation
- `router/frontend/frontend.go` - Request counting hook
- `pkg/engine/virtual_relation.go` - Extended SHOW instance output
- `pkg/config/router.go` - Config option

## Checklist

- [x] Code follows project style guidelines
- [x] Changes are backward compatible
- [x] Implementation uses thread-safe operations
- [x] References issue #1913
- [x] No changes to parser (gram.y/gram.go) required

## Additional Notes
Refactored based on initial review to avoid custom parser tokens and use `SHOW instance` extension instead.

Fixes #1913


reshke:
I can see there is changes in gram.y/gram.go which are just not necessary for me. 
I would also prefer a config setting here to turn this average aggregation off, because I dont want any new mutex on request critical path for highly-loaded installations. Something like `enable_router_rps_aggregation` (see `pkg/config`). Also, you dont need mutex for `totalRequests`  - a simple atomic counter will do. For `totalRequests`, i think we can always count and display it, even if `enable_router_rps_aggregation` is off.


I did not take thoughtful look here, just a few comment for now. 
Im also doubtful about overall design, maybe we dont need another `SHOW` target here, but simply extend `SHOW instance`>

Denchick:
@Rounakneema thanks a lot for this contribution. Just to be clear, we like the idea and the feature is definitely useful. There are just a few design adjustments we’d like to address before merging(totally agreed with @reshke here), and we’ll continue the review with specific comments.

Overall this is a strong first contribution - appreciate the effort.
We’re actively looking for collaborators, so happy to see you here 🙂

Rounakneema:
@reshke @Denchick 

I have updated the PR based on your feedback:
- **Cleaned Parser**: Removed all changes to `gram.y` and `gram.go` (and the `SHOW router_rps` command).
- **Extended `SHOW instance`**: The RPS metrics are now displayed as new columns in the existing `SHOW instance` output.
- **Atomic Counters**: `total_requests` is now tracked using a simple atomic counter (always on, minimal overhead).
- **Configurable Aggregation**: Added `router_rps_aggregation` config option (default `false`) to control the expensive sliding window calculation.

The implementation is now good and simple as requested.
Ready for review

Denchick:
Hi @Rounakneema,

I think something may have been unintentionally deleted during the merge from the `master` branch.

Please take a look here:
https://github.com/pg-sharding/spqr/pull/1992/changes#diff-8d3b46ebbb9b686794bef1e8154cbc5c8200964b26beb9e60e533ff0db4254c6L630

pkg/rps/rps.go
```diff
@@ -0,0 +1,284 @@
...
+		return
+	}
+
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.advanceWindow(now)
+
+	r.buckets[r.lastBucket]++
+	r.total++
```
reshke, line 96: Isnt r.total always equal to r.totalRequests?
you can use 
```
	newValue := atomic.AddInt64(&counter, delta)
```

to recalculate `currentRPS` below and in other places

reshke:
I would prefer to see `rps.go` inside `pkg/something`. Also I would find non-singleton design here much more welcome, so we can reuse this code to per-shard rps counting

reshke:
> I would prefer to see `rps.go` inside `pkg/something`. Also I would find non-singleton design here much more welcome, so we can reuse this code to per-shard rps counting

After putting some more thought on this, I think that we can simply move code to pkg, and have refactoring item as a separate issue. 


Having test on reload can also be split to separate effort

pkg/rps/rps.go
Rounakneema: no the r.totalRequests is lifetime total request and the r.total is the current sliding window total that degrades with time so we can't do it as it.

Rounakneema:
I have moved the rps.go in pkg rps I did that in pkg/statistics but it was colliding with router/statistics so named it rps and also added comment about the work of total and totalrequest.And as said create a new issue on which i will be proceding on it to make it non-singleton Thanks @reshke @Denchick

Rounakneema:
@Denchick @reshke corrected fmt error and update my branch as current master branch
.

pkg/rps/rps.go
```diff
@@ -0,0 +1,285 @@
...
+// InitRPSStats initializes the global RPS tracker.
+func InitRPSStats() {
+	rpsOnce.Do(func() {
+		globalRPS = NewRPSStats()
+	})
+}
+
+// InitRPSStatsWithClock initializes the global RPS tracker with a custom clock (for testing).
+// Resets the tracker if already initialized.
+func InitRPSStatsWithClock(clock Clock) {
```
reshke, line 216: This function does not seem to be used?

Rounakneema:
You are right
InitRPSStatsWithClock
 is not currently used in the codebase. It was added to support testability by allowing a mock clock to be injected, which avoids flaky time.Sleep-based tests. It will be used by unit tests for the RPS package (planned as a follow up to add this in tests ). I can remove it from this PR if you prefer and add it back when the tests are introduced.and about that one failed test it is caused by grpc race condition.

Rounakneema:
Is it done and can i close this PR and create another issue for the singleton feature and the one failing test is coming from the master branch and has nothing to infer from my code. @reshke @Denchick

Denchick:
Hey @Rounakneema,
Thank you for the contribution!

There are still a few issues and things we’ll want to refine in follow-ups, but for a first iteration this is totally fine and a solid start. The implementation direction makes sense and it’s good to see you digging into

Please keep them coming — we’d be happy to see more PRs from you :)

Rounakneema:
Hi @Denchick, @reshke thank you both for the detailed reviews and guidance throughout this process.
I really appreciate the feedback
it helped me improve the design and better understand the project’s architecture. Since this was my first contribution here the review comments were especially valuable.
I’d be happy to continue contributing. 
If there are other issues you think would be a good fit, feel free to tag me I’d be glad to take them on.

Issue #1996, opened 2026-02-05
Better `SHOW routers`


Denchick:
Expected:

```
spqr-console=> show routers;
                           router                            | status       | how much client connections | version | metadata_version
---------------------------------------------------------------+--------+-----+-----+--------
 klg-8ippclfnf9734gpf-klg-8ippclfnf9734gpf.db.yandex.net:7000 | OPENED | 700 | 2.9.1-hash | 1
 sas-6oc8u8fd9ohm2o99-sas-6oc8u8fd9ohm2o99.db.yandex.net:7000 | OPENED | 234 | 2.7.2-hash | 2
 vla-e8mtj0huv9drjgap-vla-e8mtj0huv9drjgap.db.yandex.net:7000 | OPENED | 32423 | 2.3.1-hash | 3
(3 rows)
```

(metadata version may be not implemented yet)

Actual:

```
spqr-console=> show routers;
                                 router                                 | status 
------------------------------------------------------------------------+--------
 router -> klg-8ippclfnf9734gpf-klg-8ippclfnf9734gpf.db.yandex.net:7000 | OPENED
 router -> sas-6oc8u8fd9ohm2o99-sas-6oc8u8fd9ohm2o99.db.yandex.net:7000 | OPENED
 router -> vla-e8mtj0huv9drjgap-vla-e8mtj0huv9drjgap.db.yandex.net:7000 | OPENED
(3 rows)
```

PR #1997, opened 2026-02-06
better show routers

Denchick:
https://github.com/pg-sharding/spqr/issues/1996


protos/coordinator.proto
```diff
@@ -18,6 +18,8 @@ service TopologyService {
 
 message GetRouterStatusReply {
   RouterStatus status = 1;
+  string version = 2;
+  string metadata_version = 3;
```
reshke: 64 bit int

PR #2020, opened 2026-02-11
more tests for memqdb lock and simplify fix

xelavopelk:
Thank you @EinKrebs  for quick fix my bug in PR https://github.com/pg-sharding/spqr/pull/2011 !

0) I add more unit-tests for this case in this PR
1) I had some time to reflect on your PR. I propose a more concise solution without introducing a new command type. What do you think about this idea?

### DIFF
--- a/qdb/memqdb_test.go
+++ b/qdb/memqdb_test.go
@@ -629,33 +629,7 @@ func TestTransferCreateKeyRangeQdbCommand(t *testing.T) {
 		is.NoError(err)
 		chunk, err := memqdb.CreateDistribution(ctx, qdb.NewDistribution("ds1", nil))
 		is.NoError(err)
-		err = memqdb.ExecNoTransaction(ctx, chunk)
-		is.NoError(err)
-		initKeyRange := &qdb.KeyRange{
-			KeyRangeID:     "krid1",
-			LowerBound:     [][]byte{[]byte("1")},
-			ShardID:        "sh1",
-			DistributionId: "ds1",
-			Locked:         false,
-		}
-		statements, err := memqdb.CreateKeyRange(ctx, initKeyRange)
-		is.NoError(err)
-		is.NoError(memqdb.ExecNoTransaction(ctx, statements))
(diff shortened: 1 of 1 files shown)


qdb/memqdb_test.go
```diff
@@ -619,3 +619,81 @@ func TestRestoreQDB_ValidJSON(t *testing.T) {
...
+
+	t.Run("locked false", func(t *testing.T) {
+		ctx := context.TODO()
+		is := assert.New(t)
+		memqdb, err := qdb.NewMemQDB("")
+		is.NoError(err)
+		chunk, err := memqdb.CreateDistribution(ctx, qdb.NewDistribution("ds1", nil))
+		is.NoError(err)
+		err = memqdb.ExecNoTransaction(ctx, chunk)
+		is.NoError(err)
```
EinKrebs: Please inline
```suggestion
		is.NoError(memqdb.ExecNoTransaction(ctx, chunk))
```

qdb/memqdb_test.go
```diff
@@ -619,3 +619,81 @@ func TestRestoreQDB_ValidJSON(t *testing.T) {
...
+		statements, err := memqdb.CreateKeyRange(ctx, initKeyRange)
+		is.NoError(err)
+		is.NoError(memqdb.ExecNoTransaction(ctx, statements))
+
+		lock, ok := memqdb.Locks["krid1"]
+		is.True(ok)
+		is.True(lock.TryLock())
+	})
+
+	t.Run("locked filed not defined", func(t *testing.T) {
```
EinKrebs: This test seemingly does not differ from TestTransferCreateKeyRangeQdbCommand, mb remove it?

qdb/memqdb_test.go
xelavopelk: done

qdb/memqdb_test.go
xelavopelk: ok

PR #2024, opened 2026-02-11
Fix isolation test. Add __spqr__await_task function.

reshke:


router/planner/planner.go
```diff
@@ -364,6 +365,58 @@ func PlanDistributedRelationInsert(
...
+		}
+		defer cf()
+
+		switch v := args[0].(type) {
+		case *lyx.AExprSConst:
+
+			for {
+
+				/* TODO: add CFI here*/
+				tgs, err := mgr.ListMoveTaskGroups(ctx)
```
EinKrebs: Why not use GetMoveTaskGroup here?

rkhapov (approved):

PR #2035, opened 2026-02-13
deprecate ADD keyword, use CREATE instead

Denchick:
https://github.com/pg-sharding/spqr/issues/2022


yacc/console/gram.y
reshke, line 1: without ABI changes this is good

reshke (approved):

PR #2043, opened 2026-02-15
DropKeyRange changes tranMngr

xelavopelk:
0) added dropkeyrange to transtation mngr. 
1) removed copipasta from recognition a type of gossip commands 

issue #1737

### DIFF
--- a/coordinator/pkg/clustered_coord.go
+++ b/coordinator/pkg/clustered_coord.go
@@ -2697,8 +2697,8 @@ func (qc *ClusteredCoordinator) AlterDistributedRelation(ctx context.Context, id
 
 // AlterDistributedRelationSchema changes the schema name of a relation attached to a distribution
 // TODO: unit tests
-func (qc *ClusteredCoordinator) AlterDistributedRelationSchema(ctx context.Context, id string, relName string, schemaName string) error {
-	if err := qc.Coordinator.AlterDistributedRelationSchema(ctx, id, relName, schemaName); err != nil {
+func (qc *ClusteredCoordinator) AlterDistributedRelationSchema(ctx context.Context, id string, relationName *rfqn.RelationFQN, schemaName string) error {
+	if err := qc.Coordinator.AlterDistributedRelationSchema(ctx, id, relationName, schemaName); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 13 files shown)


coordinator/pkg/clustered_coord.go
```diff
@@ -2821,7 +2821,7 @@ func (qc *ClusteredCoordinator) ExecNoTran(ctx context.Context, chunk *mtran.Met
 
 func (qc *ClusteredCoordinator) CommitTran(ctx context.Context, transaction *mtran.MetaTransaction) error {
 	for _, gossipRequest := range transaction.Operations.GossipRequests {
-		if gossipType := mtran.GetGossipRequestType(gossipRequest); gossipType == mtran.GR_UNKNOWN {
+		if _, recognized := mtran.GetGossipRequestType(gossipRequest); !recognized {
```
Denchick: I suggest a more golang-way naming (here and above):

```suggestion
		if _, ok := mtran.GetGossipRequestType(gossipRequest); !ok {
```

pkg/models/transaction/meta_transaction.go
```diff
@@ -62,9 +64,14 @@ type MetaTransactionChunk struct {
...
 // Any change in this enum must change GetGossipRequestType function
 const (
-	GR_UNKNOWN = iota
+	GR_ERROR = iota - 1
+	GR_UNKNOWN
 	GR_CreateDistributionRequest
 	GR_CreateKeyRange
+	GR_DropKeyRange
+	GR_DropKeyRangeAll
+	GR_LockKeyRange
```
Denchick: `GR_LockKeyRange` and `GR_UnLockKeyRange` are not used. Let's remove them

pkg/models/transaction/meta_transaction.go
```diff
@@ -62,9 +64,14 @@ type MetaTransactionChunk struct {
 
 // Any change in this enum must change GetGossipRequestType function
 const (
-	GR_UNKNOWN = iota
+	GR_ERROR = iota - 1
```
Denchick, line 67: seems strange

coordinator/pkg/clustered_coord.go
xelavopelk: ok

pkg/models/transaction/meta_transaction.go
xelavopelk: GR_ERROR = -1
GR_UNKNOWN = 0
--------
all significant types are greater than 0

pkg/models/transaction/meta_transaction.go
xelavopelk: ok. but they will come back soon

PR #2060, opened 2026-02-18
Revert to default max etcd retries and increase backoff

EinKrebs:
Default max retries number was actually 100, so restore default setting.


EinKrebs:
Closes #2042, as timeout errors are reqtryable by the client

PR #2061, opened 2026-02-18
Add DROP MOVE TASK command

EinKrebs:


pkg/coord/adapter.go
```diff
@@ -983,6 +983,38 @@ func (a *Adapter) ListMoveTasks(ctx context.Context) (map[string]*tasks.MoveTask
...
+
+// RemoveMoveTask removes move task with given ID from the system.
+//
+// Parameters:
+// - ctx (context.Context): The context for the request.
+// - id  (string): The ID of the move task to remove.
+//
+// Returns:
+// - error: An error if the retrieval of tasks fails, otherwise nil.
+func (a *Adapter) RemoveMoveTask(ctx context.Context, id string) error {
```
Denchick, line 1012: It is strange that the command called DROP but inner methods are Remove

PR #2065, opened 2026-02-18
Refactor batch move & retry task group to avoid code duplication

reshke:


coordinator/pkg/clustered_coord.go
```diff
@@ -1197,6 +1197,42 @@ func (qc *ClusteredCoordinator) checkKeyRangeMove(ctx context.Context, req *kr.B
 	return nil
 }
 
+/*
+* Workhorse for all move data operations
+ */
+func (qc *ClusteredCoordinator) executeMoveInternal(
+	ctx context.Context,
+	taskGroup *tasks.MoveTaskGroup,
+	invalidateTG bool,
```
EinKrebs, line 1206: We can invalidate all the time

PR #2072, opened 2026-02-20
Key range versioning basics

EinKrebs:


Denchick (approved):

xelavopelk:
What do we want to get by implementing the key range version?

PR #2101, opened 2026-02-25
Put coordinator address in task group lock & status message

EinKrebs:


reshke (approved):

qdb/memqdb.go
```diff
@@ -1850,7 +1850,7 @@ func (q *MemQDB) TXStatus(gid string) string {
 //                               TASK GROUP STATE
 // ==============================================================================
 
-func (q *MemQDB) TryTaskGroupLock(ctx context.Context, tgId string) error {
+func (q *MemQDB) TryTaskGroupLock(ctx context.Context, tgId string, holder string) error {
```
Denchick, line 1853: leaser sounds better
anyway

EinKrebs: It's lease holder, or lessee, not leaser. I think we should keep `holder`

PR #2105, opened 2026-02-26
Persist shard TLS config and add UpdateShard RPC with hardened router sync

Ledatu:
## Problem

When `manage_shards_by_coordinator=true`, shard TLS configuration (`sslmode`, `cert_file`, `key_file`, `root_cert_file`) is silently dropped during the coordinator → etcd → router pipeline. Routers end up connecting to SSL-enabled PostgreSQL shards in plaintext.

This makes `manage_shards_by_coordinator` unusable in any deployment where shards are not on a trusted local network, since there is no way to enforce encrypted connections.

The root cause is that the conversion functions between `config.Shard`, `qdb.Shard`, and `proto.Shard` did not carry the TLS fields, and `qdb.NewShard()` only accepted `id` and `hosts`.

## Fix

**Proto layer** — add `TLSConfig` message and `tls` field to `Shard` in `protos/shard.proto`.

**QDB layer** — add standalone `TLSConfig` struct and `TLS` field to `qdb.Shard` (with `omitempty` for backward compat).

**Config layer** — add `TLS` field to `config.ShardConnect` so `shard_data.yaml` can specify TLS per shard.

**Topology layer** — add four conversion helpers (`tlsConfigToProto`, `tlsConfigFromProto`, `tlsConfigToDB`, `tlsConfigFromDB`) and plumb TLS through `DataShardToDB`, `DataShardFromDB`, `DataShardToProto`, `DataShardFromProto`.

**Coordinator** — fix `AddDataShard` to use `DataShardToDB` instead of `NewShard` (which dropped TLS). Fix `RunCoordinator` to persist TLS when loading `shard_data.yaml`.

## Tests

- Unit tests for all conversion paths including nil handling and round-trips (`pkg/models/topology/shards_test.go`)
- Docker-based integration test verifying TLS is used end-to-end: coordinator persists `sslmode=require` in etcd, router loads it, and `pg_stat_ssl` confirms TLS connections to shards (`test/tls-integration/`)

### DIFF
--- a/pkg/coord/clocal.go
+++ b/pkg/coord/clocal.go
@@ -28,8 +28,7 @@ type LocalInstanceMetadataMgr struct {
 	cache *cache.SchemaCache
 
 	updateShardsMapping bool
-	shardMapping        map[string]*config.Shard
-	shardMappingMutex   sync.Mutex
+	shardMapping        sync.Map
 
 	poolShardHosts shard.ShardHostIterator
 }
@@ -218,9 +217,7 @@ func (lc *LocalInstanceMetadataMgr) AddDataShard(ctx context.Context, ds *topolo
 		Msg("adding datashard node in local coordinator")
 
 	if lc.updateShardsMapping {
-		lc.shardMappingMutex.Lock()
-		lc.shardMapping[ds.ID] = ds.Cfg
-		lc.shardMappingMutex.Unlock()
+		lc.shardMapping.Store(ds.ID, ds.Cfg)
 	}
 	return lc.Coordinator.AddDataShard(ctx, ds)
 }
@@ -234,35 +231,33 @@ func (lc *LocalInstanceMetadataMgr) UpdateShard(ctx context.Context, ds *topolog
 		return err
 	}
 	if lc.updateShardsMapping {
-		lc.shardMappingMutex.Lock()
-		lc.shardMapping[ds.ID] = ds.Cfg
-		lc.shardMappingMutex.Unlock()
+		lc.shardMapping.Store(ds.ID, ds.Cfg)
 	}
-	lc.invalidatePoolsForShard(ds.ID)
-	return nil
+	return lc.invalidatePoolsForShard(ds.ID)
 }
 
-func (lc *LocalInstanceMetadataMgr) invalidatePoolsForShard(shardID string) {
+func (lc *LocalInstanceMetadataMgr) invalidatePoolsForShard(shardID string) error {
 	if lc.poolShardHosts == nil {
-		return
+		return nil
 	}
-	if err := lc.poolShardHosts.ForEach(func(sh shard.ShardHostCtl) error {
+
+	err := lc.poolShardHosts.ForEach(func(sh shard.ShardHostCtl) error {
 		if sh.ShardKeyName() == shardID {
 			sh.MarkStale()
 		}
 		return nil
-	}); err != nil {
+	})
+	if err != nil {
 		spqrlog.Zero.Error().Err(err).
 			Str("shard", shardID).
 			Msg("pool invalidation: error iterating connections")
 	}
+	return err
 }
 
 func (lc *LocalInstanceMetadataMgr) DropShard(ctx context.Context, shardId string) error {
 	if lc.updateShardsMapping {
-		lc.shardMappingMutex.Lock()
-		delete(lc.shardMapping, shardId)
-		lc.shardMappingMutex.Unlock()
+		lc.shardMapping.Delete(shardId)
 	}
 	return lc.qdb.DropShard(ctx, shardId)
 }
@@ -490,13 +485,20 @@ func (lc *LocalInstanceMetadataMgr) SyncReferenceRelations(ctx context.Context, 
 //
 // Returns:
 // - meta.EntityMgr: The newly created LocalCoordinator instance.
-func NewLocalInstanceMetadataMgr(db qdb.XQDB, d qdb.DCStateKeeper, cache *cache.SchemaCache, shardMapping map[string]*config.Shard, updateShardsMapping bool, poolShardHosts shard.ShardHostIterator) meta.EntityMgr {
-	return &LocalInstanceMetadataMgr{
+func NewLocalInstanceMetadataMgr(db qdb.XQDB, d qdb.DCStateKeeper, cache *cache.SchemaCache,
+	shardMapping map[string]*config.Shard, updateShardsMapping bool, poolShardHosts shard.ShardHostIterator) meta.EntityMgr {
+
+	lc := &LocalInstanceMetadataMgr{
 		Coordinator:         NewCoordinator(db, d),
 		cache:               cache,
-		shardMapping:        shardMapping,
-		shardMappingMutex:   sync.Mutex{},
+		shardMapping:        sync.Map{},
 		updateShardsMapping: updateShardsMapping,
 		poolShardHosts:      poolShardHosts,
 	}
+
+	for k, v := range shardMapping {
+		lc.shardMapping.Store(k, v)
+	}
+
+	return lc
 }
--- a/router/instance/instance.go
+++ b/router/instance/instance.go
@@ -125,6 +125,7 @@ func NewRouter(ctx context.Context, ns string) (*InstanceImpl, error) {
 		config.RouterConfig().ManageShardsByCoordinator,
 		rr,
 	)
+
 	var seqMngr sequences.SequenceMgr = lc
 	idRangeSize := config.RouterConfig().IdentityRangeSize
 	var identityMgr planner.IdentityRouterCache = planner.NewIdentityRouterCache(idRangeSize, &seqMngr)


Denchick:
Thank you for this PR! We really appreciate your contribution and think this is a good idea. We'll review it shortly.

coordinator/pkg/clustered_coord.go
```diff
@@ -595,7 +595,16 @@ func (qc *ClusteredCoordinator) RunCoordinator(ctx context.Context, initialRoute
 						Msg("already exists. creating shard skipped")
 					continue
 				}
-				if err := qc.db.AddShard(context.TODO(), qdb.NewShard(id, cfg.Hosts)); err != nil {
+				newShard := qdb.NewShard(id, cfg.Hosts)
+				if cfg.TLS != nil {
+					newShard.TLS = &qdb.TLSConfig{
```
Denchick: tlsConfigToDB?

Ledatu:
Thanks for feedback. I will try to rewrite the tests myself in the next few days. 

upd: Turned out to be an interesting one, hope you'll like it when im done :)

Ledatu:
Started with `CREATE SHARD` and `ALTER SHARD` for TLS fields, but the feature tests showed TLS config was being dropped during the coordinator→router sync cycle — the shard model and proto didn't carry TLS, and there was no way to update a shard on a router without drop + re-add. That pushed me to implement the full persistence path end-to-end.

## Changes

**TLS persistence:** Extended `Shard` proto with `TLSConfig`, added `UpdateShard` to qdb/etcd/memqdb, wired TLS through all shard model conversions, added `ALTER SHARD` grammar with parser tests.

**UpdateShard RPC:** New RPC on `ShardService` used by both `ALTER SHARD` and `SyncRouterMetadata`. Includes a `fallbackDropAndAdd` path for older routers that don't support it, with snapshot + rollback.

**Coordinator hardening:**

- `traverseRoutersAll` — non-fast-fail traversal so healthy routers get updated even when one is down
- `fallbackDropAndAdd` — aborts before destructive drop if snapshot fails unexpectedly
- `SyncRouterMetadata` — shard loops continue on per-shard errors instead of aborting on first failure

**Tests:** TLS shard feature tests, 11 fallback unit tests, parser tests for new syntax.

Would like to hear your thoughts on this. For my use case I'd really like to be able to rotate certs on shards without downtime, and to spin up backup replicas with the data already there and then just swap the hosts on the fly via `ALTER SHARD`.

Ledatu:
Merged upstream for USE_PGXS fix

coordinator/pkg/clustered_coord.go
```diff
@@ -601,7 +602,9 @@ func (qc *ClusteredCoordinator) RunCoordinator(ctx context.Context, initialRoute
 						Msg("already exists. creating shard skipped")
 					continue
 				}
-				if err := qc.db.AddShard(context.TODO(), qdb.NewShard(id, cfg.Hosts)); err != nil {
+				newShard := qdb.NewShard(id, cfg.Hosts)
```
diPhantxm: maybe pass TLS config into qdb.NewShard?

coordinator/pkg/clustered_coord.go
```diff
@@ -2581,6 +2679,119 @@ func (qc *ClusteredCoordinator) AddDataShard(ctx context.Context, shard *topolog
...
+	// 2. Drop the shard.
+	if _, err := cl.DropShard(ctx, &proto.DropShardRequest{Id: shardID}); err != nil {
+		return err
+	}
+
+	// 3. Re-add with the new config, retrying up to 3 times for transient errors.
+	newShardProto := topology.DataShardToProto(shard)
+	var addErr error
+	const (
+		maxAddAttempts  = 3
```
diPhantxm: Can we use retry package? Like we do [here](https://github.com/pg-sharding/spqr/blob/master/coordinator/pkg/clustered_coord.go#L2166)

coordinator/pkg/clustered_coord.go
```diff
@@ -2581,6 +2679,119 @@ func (qc *ClusteredCoordinator) AddDataShard(ctx context.Context, shard *topolog
...
+	spqrlog.Zero.Error().Err(addErr).
+		Str("shard", shardID).
+		Msg("fallback re-add exhausted retries; shard is missing from router until next reconciliation")
+
+	// 5. Attempt rollback to the previous config if we have a snapshot.
+	if oldShard != nil {
+		spqrlog.Zero.Warn().
+			Str("shard", shardID).
+			Msg("attempting rollback to previous shard config on router")
+		if _, rbErr := cl.AddDataShard(ctx, &proto.AddShardRequest{Shard: oldShard}); rbErr != nil {
```
diPhantxm: If we fail on adding shard back on router, when will we retry?

I am thinking about this approach to re-add shards on router that does not support Update. I came up with some questions: 1. what if we fail to re-add? we dropped the shard, but didn't manage to add it back with new settings. 2. With `manage_shards_by_coordinator: true` on router it will probably drop some connections

Maybe it's better to say that router does not support update? Update all routers or none? idk

pkg/coord/clocal.go
```diff
@@ -222,6 +230,25 @@ func (lc *LocalInstanceMetadataMgr) AddDataShard(ctx context.Context, ds *topolo
...
+	spqrlog.Zero.Info().
+		Str("node", ds.ID).
+		Msg("updating datashard node in local coordinator")
+
+	if lc.updateShardsMapping {
+		lc.shardMappingMutex.Lock()
+		lc.shardMapping[ds.ID] = ds.Cfg
+		lc.shardMappingMutex.Unlock()
+	}
+	if err := lc.Coordinator.UpdateShard(ctx, ds); err != nil {
```
diPhantxm: What if we fail to update shard in memqdb here, but we already updated shard config in mapping above?

coordinator/pkg/clustered_coord.go
```diff
@@ -2581,6 +2679,119 @@ func (qc *ClusteredCoordinator) AddDataShard(ctx context.Context, shard *topolog
...
+				Str("shard", shardID).
+				Msg("rollback succeeded; shard restored with previous config on router")
+		}
+	}
+
+	return addErr
+}
+
+func (qc *ClusteredCoordinator) UpdateShard(ctx context.Context, shard *topology.DataShard) error {
+	if err := qc.db.UpdateShard(ctx, topology.DataShardToDB(shard)); err != nil {
```
diPhantxm: I think updating shard in qdb should be moved into Coordinator, since both memqdb and etcdqdb should update shard, this is common logic

pkg/meta/meta.go
```diff
@@ -582,10 +582,23 @@ func ProcessCreate(ctx context.Context, astmt spqrparser.Statement, mngr EntityM
 		}
 		return tts, nil
 	case *spqrparser.ShardDefinition:
-		dataShard := topology.NewDataShard(stmt.Id, &config.Shard{
+		if stmt.SslMode != "" && !topology.ValidSslMode(stmt.SslMode) {
```
diPhantxm: should we move all possible values into grammar? We could make parser do the validation

pkg/models/topology/shards.go
```diff
@@ -80,12 +132,68 @@ func DataShardFromDB(shard *qdb.Shard) *DataShard {
...
+	return a.SslMode == b.SslMode &&
+		a.CertFile == b.CertFile &&
+		a.KeyFile == b.KeyFile &&
+		a.RootCertFile == b.RootCertFile
+}
+
+// ShardConfigEqual reports whether two DataShards have identical configuration
+// (hosts and TLS). This is used by SyncRouterMetadata to detect shards
+// that exist on both the coordinator and the router but have drifted.
+func ShardConfigEqual(a, b *DataShard) bool {
```
diPhantxm, line 155: Why don't we just use reflect.DeepEqual?

router/instance/instance.go
```diff
@@ -129,6 +130,24 @@ func NewRouter(ctx context.Context, ns string) (*InstanceImpl, error) {
 	// request router
 	rr := rulerouter.NewRouter(frTLS, config.RouterConfig(), notifier)
 
+	type poolInvalidatable interface {
+		SetPoolInvalidator(func(string))
+	}
+	if pi, ok := lc.(poolInvalidatable); ok {
```
diPhantxm: It seems redundant to me, you added `SetPoolInvalidator` method to LocalCoordinator, but now check if it implements `SetPoolInvalidator`. Can we just remove this interface and type assertion?

And I also could not find any place where you could change pool invalidator function, but this. So if we can omit this interface, I believe we can remove `SetPoolInvalidator` and just specify `InvalidatePool` function

pkg/models/topology/shards.go
Ledatu: We can’t `DeepEqual` the whole `DataShard/Shard` because host config has multiple raw representations. `RawHosts` may include AZ suffixes like `host1:5432:spb-1a`, while `Hosts()` normalizes that to `host1:5432`, so coordinator and router can differ textually without actually drifting. `DeepEqual` would also include cached parsed fields on `config.Shard`, which are not part of the logical config. `TLS` has no such normalization, so `DeepEqual` is fine there and is now used, thanks for pointing it out

diPhantxm:
@Ledatu soft ping
Hi! What is status of this PR? We really need this feature done. Please reach out if you still want to work on this issue

Ledatu:
@diPhantxm Hello! I'd like to help with it, just want to get the opinion of maintainers - if there will be reviews I'll try to quickly resolve anything y'all find or add anything that's needed here too. Tomorrow I'll rebase it to master so there is no conflicts, and will wait for reviews, how does this sound?

diPhantxm:
> @diPhantxm Hello! I'd like to help with it, just want to get the opinion of maintainers - if there will be reviews I'll try to quickly resolve anything y'all find or add anything that's needed here too. Tomorrow I'll rebase it to master so there is no conflicts, and will wait for reviews, how does this sound?

Sounds great! I will check it asap.
@Denchick please review this pr too

pkg/coord/clocal.go
```diff
@@ -29,6 +29,8 @@ type LocalInstanceMetadataMgr struct {
 	updateShardsMapping bool
 	shardMapping        map[string]*config.Shard
 	shardMappingMutex   sync.Mutex
+
+	poolInvalidator func(shardID string)
```
diPhantxm: If you store poolInvalidtor as structure field, it's supposed to change somewhere. Or if it's passed in constructor, you may want to use different function for tests. But I don't find any of these cases in code. So let's just make it a structure function?

router/instance/instance.go
```diff
@@ -81,8 +82,10 @@ func NewRouter(ctx context.Context, ns string) (*InstanceImpl, error) {
 
 	cache := cache.NewSchemaCache(config.RouterConfig().ShardMapping, config.RouterConfig().SchemaCacheBackendRule)
 
-	lc := coord.NewLocalInstanceMetadataMgr(db, db, cache, config.RouterConfig().ShardMapping, config.RouterConfig().ManageShardsByCoordinator)
-
+	d, err := qdb.NewDataPlaneTwoPhaseStateKeeper("mem")
```
diPhantxm: ??
I guess it's rebase issues

coordinator/pkg/clustered_coord.go
```diff
@@ -748,6 +749,50 @@ func (qc *ClusteredCoordinator) traverseRouters(ctx context.Context, cb func(cc
 	return nil
 }
 
+// traverseRoutersBestEffort is like traverseRouters but does not fast-fail.
+// It attempts the callback on every opened router and returns a combined
+// error (via errors.Join) if any router(s) failed, so that healthy routers
+// are not blocked by a single unhealthy one.  Use this for propagation
+// operations (e.g. UpdateShard) where partial success is better than
+// all-or-nothing.
+func (qc *ClusteredCoordinator) traverseRoutersBestEffort(ctx context.Context, cb func(cc *grpc.ClientConn) error) error {
```
diPhantxm: I am not sure, but do we really need to traverse all the routers and ignore error here? I mean in other operations we fail on first router and we live with that some routers now have different metadata that the others. imo we need to keep the same behavior. opinions?

router/instance/instance.go
Ledatu: yeah seems like it. shame on me

coordinator/pkg/clustered_coord.go
Ledatu: Agreed. I’ve reverted this to the standard fail-fast behavior to ensure consistency with the rest of the coordinator operations. It’s better not to introduce a one-off error-handling pattern here :)

pkg/coord/clocal.go
Ledatu: Instead of a standalone function, I’ve updated the struct to take a `shard.ShardHostIterator`. This allows the `LocalInstanceMetadataMgr` to handle the invalidation logic internally using the provided pool hosts, which is cleaner than passing a naked callback. I've updated the constructor accordingly

reshke:
Will take care of `yacc` diffs here and give a review.

reshke:
So, okay.
Lets split pr into reviewable and revertible parts. It will be no-brainer to move all new yacc & proto primitives to another PR and merge it into master. Then, rebase this on new master and we will continue.

Ledatu:
Moved all new yacc and proto primitives into separate PR #2303

reshke:
I really hate new field to `NewLocalInstanceMetadataMgr`, but I decided to fight this in separate refactoring PR.
Added few kibitzing, will merge

PR #2120, opened 2026-03-01
converting drop key range command into 2 phase execution

xelavopelk:
converting drop key range command into 2 phase execution
issue #1737


qdb/etcdqdb.go
```diff
@@ -2805,7 +2805,7 @@ func (q *EtcdQDB) CommitTransaction(ctx context.Context, transaction *QdbTransac
 		Commit()
 
 	if err != nil {
-		return fmt.Errorf("failed to commit transaction: %s", transaction.Id())
+		return fmt.Errorf("failed to commit transaction: %s, error: %w", transaction.Id(), err)
```
Denchick:
```suggestion
		return fmt.Errorf("failed to commit transaction %s: %w", transaction.Id(), err)
```

pkg/meta/meta.go
```diff
@@ -172,9 +171,10 @@ func processDrop(ctx context.Context,
 		if len(krs) != 0 && !isCascade {
 			return nil, spqrerror.Newf(spqrerror.SPQR_INVALID_REQUEST, "cannot drop distribution %s because other objects depend on it\nHINT: Use DROP ... CASCADE to drop the dependent objects too.", stmt.ID)
 		}
-
+		// TODO: need pack to batch
```
Denchick, line 174: Do you have any ideas potentially how?

xelavopelk: The infrastructure for this is provided by TranEntityManager. For safe handling of batches, it is necessary to check the package size. Etcd has a default limit of 128 commands per packet.

xelavopelk: this limit is in issue too
image

PR #2131, opened 2026-03-02
Manage shard_data in qdb

diPhantxm:


Ledatu:
Hey! I've been working on a parallel change that persists TLS config for shards #2105  and noticed significant overlap with this PR.

I took a different approach to shard mutation: a single `UpdateShard(ctx, *Shard)` that does full-config replace (read-modify-write), rather than per-field methods like `AlterShardHosts` / `AlterShardOptions` through every layer.

Main reasons:

- Adding new fields (TLS, options, etc.) needs zero interface changes — just add to the `Shard` model
- `SyncRouterMetadata` already does full-state reconciliation, so the router RPC is naturally a full replace
- Much less surface area: one qdb method, one RPC, one mock instead of N per field group

Your `Options` field and `loadShardsConnectionData()` would work identically with this pattern — the only difference is swapping the two alter methods for one `UpdateShard`.

Also noticed a small bug: `MemQDB.AlterShardHosts` and `AlterShardOptions` check `if ok` but return `"already exists"` — should be `if !ok`.

Happy to coordinate so we don't duplicate work!

diPhantxm:
> Hey! I've been working on a parallel change that persists TLS config for shards #2105 and noticed significant overlap with this PR.
> 
> I took a different approach to shard mutation: a single `UpdateShard(ctx, *Shard)` that does full-config replace (read-modify-write), rather than per-field methods like `AlterShardHosts` / `AlterShardOptions` through every layer.
> 
> Main reasons:
> 
> * Adding new fields (TLS, options, etc.) needs zero interface changes — just add to the `Shard` model
> * `SyncRouterMetadata` already does full-state reconciliation, so the router RPC is naturally a full replace
> * Much less surface area: one qdb method, one RPC, one mock instead of N per field group
> 
> Your `Options` field and `loadShardsConnectionData()` would work identically with this pattern — the only difference is swapping the two alter methods for one `UpdateShard`.
> 
> Also noticed a small bug: `MemQDB.AlterShardHosts` and `AlterShardOptions` check `if ok` but return `"already exists"` — should be `if !ok`.
> 
> Happy to coordinate so we don't duplicate work!

Hey! Thank you for notification! 
That's a great approach! 

I will then wait for resolution of your PR and then continue with mine!
And thank you for spotting a bug, I really missed that :) This pr lacks of some tests...

Denchick:
Thanks for the PR, this looks like a useful contribution. However, since there hasn’t been any activity here for a long time and the branch now has many conflicts with the current codebase, I’d prefer to close this PR for now.

If someone wants to continue this work later, it would probably be better to reopen it as a fresh PR based on the current main branch.

PR #2144, opened 2026-03-03
Reorder & improve operations in RunCoordiantor

EinKrebs:
1. Start watching routers & task groups before finishing key range moves
2. Finish key range moves in parallel

### DIFF
--- a/coordinator/pkg/clustered_coord.go
+++ b/coordinator/pkg/clustered_coord.go
@@ -604,7 +604,6 @@ func (qc *ClusteredCoordinator) RunCoordinator(ctx context.Context, initialRoute
 		}
 	}
 
-	go qc.watchRouters(context.TODO())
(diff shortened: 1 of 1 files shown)


coordinator/pkg/clustered_coord.go
```diff
@@ -604,6 +604,9 @@ func (qc *ClusteredCoordinator) RunCoordinator(ctx context.Context, initialRoute
 		}
 	}
 
+	go qc.watchRouters(context.TODO())
```
Denchick: // TODO graceful shutdown

)))))))))

reshke (approved):

coordinator/pkg/clustered_coord.go
EinKrebs: wdym? we don't need a shutdown here, it's meant to run indefinitely

PR #2167, opened 2026-03-08
qdb meta transaction infra for CreateSequence

xelavopelk:
qdb meta transaction infra for CreateSequence
issue #1737


Denchick:
Maybe it is a good idea to combine #2167 and #2181 together?

xelavopelk:
> Maybe it is a good idea to combine #2167 and #2181 together?

2181 have been merged

pkg/coord/clocal.go
```diff
@@ -77,9 +77,19 @@ func (lc *LocalInstanceMetadataMgr) AlterDistributedRelation(ctx context.Context
...
+			return err
+		}
+		if ok {
+			return fmt.Errorf("the sequence %s already exists", SeqName)
+		}
+		statements, err := lc.qdb.CreateSequence(ctx, SeqName, 0)
+		if err != nil {
+			return err
+		}
+		if lc.qdb.ExecNoTransaction(ctx, statements) != nil {
```
Denchick:
```suggestion
		err := lc.qdb.ExecNoTransaction(ctx, statements)
		if err != nil {
```

xelavopelk: ok

Issue #1966, opened 2026-01-26
Feature: Improve Error Messages


Denchick:
**Current:**
```
ERROR: rpc error: code = Unknown desc = shard "shard2" not found
```

**Proposed:**
```
ERROR: Shard "shard2" not found.

Available shards: shard1
Hint: Run 'SHOW SHARDS' to see all configured shards.
      Run 'ADD SHARD shard2 WITH HOSTS ...' to create it.
```

Related to: #1887

Denchick:
We discussed this with @reshke and agreed that this is an important problem to solve

Minetoff:
I'm taking it to work

Denchick:
The shard-not-found part has been addressed in #2175: `GetShard` now returns a typed `SpqrError` with code, message, and a `SHOW shards` hint, and `SpqrError` metadata is preserved through gRPC transport.

The remaining work is to audit other error paths across the codebase and apply the same pattern — replacing raw/wrapped error strings with structured, user-actionable `SpqrError` messages (e.g., distribution not found, key range, connection errors, etc.).

Denchick:
## Audit: raw `fmt.Errorf` that should use `SpqrError`

After #2175 established the pattern (`ShardNotFound` constructor + gRPC metadata preservation), here are all remaining user-facing errors that should follow the same approach.

### QDB layer (`SPQR_NO_DATASHARD` / `SPQR_OBJECT_NOT_EXIST`)

**`qdb/memqdb.go`**
- [`L714`](https://github.com/pg-sharding/spqr/blob/166b57d2/qdb/memqdb.go#L714) — `router "%s" not found`
- [`L751`](https://github.com/pg-sharding/spqr/blob/166b57d2/qdb/memqdb.go#L751) — `shard with id %s already exists`
- [`L1107`](https://github.com/pg-sharding/spqr/blob/166b57d2/qdb/memqdb.go#L1107) — `reference relation "%s" not found`
- [`L1211`](https://github.com/pg-sharding/spqr/blob/166b57d2/qdb/memqdb.go#L1211) — `distribution "%s" not found` (unique index creation)
- [`L1235`](https://github.com/pg-sharding/spqr/blob/166b57d2/qdb/memqdb.go#L1235) — `unique index "%s" not found`
- [`L1300`](https://github.com/pg-sharding/spqr/blob/166b57d2/qdb/memqdb.go#L1300) — `task group "%s" already exists`
[...]

PR #2175, opened 2026-03-12
Improve shard-not-found errors in console.

Minetoff:
Provide actionable shard-not-found messages for key-range validation with available shard list and hints, and clean gRPC wrappers before emitting console errors so users see the underlying problem directly.

### DIFF
--- a/pkg/clientinteractor/interactor_test.go
+++ b/pkg/clientinteractor/interactor_test.go
@@ -590,11 +590,10 @@ func TestReportErrorIncludesHintFromSpqrError(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	ca := mockcl.NewMockRouterClient(ctrl)
 
-	spErr := spqrerror.NewWithHint(
+	spErr := spqrerror.New(
 		spqrerror.SPQR_NO_DATASHARD,
 		"Shard \"shard2\" not found.",
-		"Run 'SHOW shards' to see all configured shards.",
-	)
+	).Hint("Run 'SHOW shards' to see all configured shards.")
 
 	gomock.InOrder(
 		ca.EXPECT().ReplyErrMsgPure(spErr),
--- a/pkg/clientinteractor/interactor_test.go
+++ b/pkg/clientinteractor/interactor_test.go
@@ -24,6 +24,8 @@ import (
 	"github.com/pg-sharding/spqr/router/client"
 	mockcl "github.com/pg-sharding/spqr/router/mock/client"
 	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/pg-sharding/spqr/pkg/clientinteractor"
 	spqrparser "github.com/pg-sharding/spqr/yacc/console"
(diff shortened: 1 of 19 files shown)


test/regress/tests/console/expected/add.out
```diff
@@ -38,7 +38,11 @@ SHOW key_ranges;
 (1 row)
 
 CREATE KEY RANGE krid2 FROM 33 ROUTE TO nonexistentshard FOR DISTRIBUTION ds1;
-ERROR:  unknown shard nonexistentshard
+ERROR:  Shard "nonexistentshard" not found.
+
+Available shards: sh1, sh2, sh3, sh4
+Hint: Run 'SHOW SHARDS' to see all configured shards.
+      Run 'ADD SHARD nonexistentshard WITH HOSTS ...' to create it.
```
Denchick:
```suggestion
ERROR:  Shard "nonexistentshard" not found.
HINT: Run 'SHOW shards' to see all configured shards.
```

test/regress/tests/console/expected/add_1.out
```diff
@@ -38,7 +38,11 @@ SHOW key_ranges;
 (1 row)
 
 CREATE KEY RANGE krid2 FROM 33 ROUTE TO nonexistentshard FOR DISTRIBUTION ds1;
-ERROR:  unknown shard nonexistentshard
+ERROR:  Shard "nonexistentshard" not found.
```
Denchick, line 41: Seems like test/regress/tests/console/expected/add.out and test/regress/tests/console/expected/add_1.out are the same

pkg/meta/key_range_meta.go
```diff
@@ -18,6 +20,43 @@ const (
...
+		return nil
+	}
+
+	sort.Strings(shardIDs)
+	availableShards := "none"
+	if len(shardIDs) > 0 {
+		availableShards = strings.Join(shardIDs, ", ")
+	}
+
+	return spqrerror.Newf(
```
Denchick: Hint field is a part of postgresql protocol, you should use `spqrerror.NewWithHint` instead

pkg/clientinteractor/interactor_test.go
```diff
@@ -584,6 +586,25 @@ func TestBackendConnectionsGroupByFail(t *testing.T) {
 	assert.ErrorContains(err, "failed to resolve 'someColumn' column offset")
 }
 
+func TestReportErrorCleansGrpcPrefix(t *testing.T) {
```
Denchick: I'm not sure whether we indeed need this test or not. I think we could safely remove it

pkg/meta/key_range_meta.go
```diff
@@ -18,6 +20,43 @@ const (
 	LockRetryStep = 500 * time.Millisecond
 )
 
+func validateTargetShardExists(ctx context.Context, mngr EntityMgrReader, shardID string) error {
+	shards, err := mngr.ListShards(ctx)
```
Denchick: I suggest you to rewrite the error message to the following format:

```
ERROR:  Shard "nonexistentshard" not found.
HINT: Run 'SHOW shards' to see all configured shards.
```

Then you may use `mngr.GetShard` instead of `ListShards`, it will be more efficient and more postgresql-way.

Denchick (changes requested):
Seems like this PR is related to https://github.com/pg-sharding/spqr/issues/1966

pkg/meta/key_range_meta.go
```diff
@@ -18,17 +20,37 @@ const (
...
+	if errors.As(err, &spErr) && spErr.ErrorCode == spqrerror.SPQR_NO_DATASHARD {
+		return spqrerror.NewWithHint(
+			spqrerror.SPQR_NO_DATASHARD,
+			fmt.Sprintf("Shard %q not found.", shardID),
+			"Run 'SHOW shards' to see all configured shards.",
+		)
+	}
+
+	cleanErr := strings.ToLower(spqrerror.CleanGrpcError(err).Error())
+	if strings.Contains(cleanErr, "unknown shard") || strings.Contains(cleanErr, "shard") && strings.Contains(cleanErr, "not found") {
```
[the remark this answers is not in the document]
Minetoff: Done, made the condition explicit for readability

Denchick (changes requested):
Thanks for working on this. The user-facing error format looks right, but I think we should avoid fixing it by parsing/re-writing error strings in `meta`.

`SpqrError` already has `code/message/hint/detail` fields, so I’d suggest defining this as a typed domain error at the source instead:

```go
spqrerror.ShardNotFound(shardID)
```

That constructor can return `SPQR_NO_DATASHARD` with message `Shard "<id>" not found.` and the `SHOW shards` hint.

Then `qdb.GetShard` should return that error directly, and `ValidateKeyRangeForCreate/Modify` can just return `mngr.GetShard(...)` errors unchanged. This avoids fragile string matching and gives us a reusable pattern for future user-facing errors.

Separately, we should improve `CleanGrpcError` / gRPC transport so `SpqrError` metadata is preserved instead of being collapsed into a plain string.

Denchick (approved):

Denchick:

PR #2177, opened 2026-03-13
fix: deep-copy pgproto3 message byte slices to prevent cross-client data contamination

Ledatu:
## Summary

`pgproto3.Backend.Receive()` returns flyweight message structs whose `[][]byte` fields (`Bind.Parameters`, `FunctionCall.Arguments`, `DataRow.Values`) are sub-slices of a shared `chunkReader` buffer backed by a global `sync.Pool` (`iobufpool`). The pgproto3 docs explicitly state:

> "The returned message is only valid until the next call to Receive."

SPQR's `frontend.ProcessMessage` performs a shallow struct copy (`cpQ := *q`) before storing these messages in `xBuf` for deferred processing at Sync time. This copies the struct but **not** the underlying byte data — the `[][]byte` elements still point into the shared read buffer. Between the Bind `Receive()` and the eventual `ProcessExtendedBuffer`, subsequent `Receive()` calls can trigger `chunkReader` buffer recycling (`iobufpool.Put` / `iobufpool.Get`), allowing a completely different client goroutine to overwrite the buffer contents.

## Bug chain

```
pgproto3.Bind.Decode(src):
    dst.Parameters[i] = src[rp : rp+msgSize]     // sub-slice, NOT a copy

chunkReader.Next():
    if r.rp == r.wp {
        iobufpool.Put(r.buf)                       // recycle to global sync.Pool
        r.buf = iobufpool.Get(r.minBufSize)        // may come from ANY goroutine
    }

frontend.ProcessMessage():
    cpQ := *q    // shallow copy — Parameters still references the recycled buffer
    rst.AddExtendedProtocMessage(q)                // stored until Sync arrives
```

## Observed symptoms in production

Under concurrent load (multiple application consumers hitting the same router instance), query parameters were cross-contaminated with data from unrelated clients:

- **Integer parameter received Cyrillic text**: a `SELECT ... WHERE nmId IN ($1, $2)` with integer params received `"подч"` (a fragment of a Russian word from another client's INSERT)
- **Integer parameter received URL fragment**: same query type received `"61119/76"` (a path segment from another client's query)
- **Raw bytes in text parameters**: `0xbe` appearing in what should have been UTF-8 text

Retries did not help because the corruption was in the router's buffer pool, not in the connection. Every new query through the same SPQR router instance hit the same poisoned buffer pool.

## Root cause diagram

```
Client A: Bind(nmId=12345)
  → cl.Receive() → Backend.Receive()
    → chunkReader.Next() returns sub-slice of buf
      → Bind.Decode: Parameters[i] = src[rp:rp+msgSize]
        → frontend.go: cpQ := *q (shallow copy)
          → xBuf stores Bind; Parameters still points into buf

Client A: next cl.Receive() for Execute/Sync
  → chunkReader: rp == wp → iobufpool.Put(buf) → returned to global sync.Pool

Client B goroutine (concurrent):
  → iobufpool.Get() → receives the SAME buf
    → Client B fills buf with its own data (e.g. Cyrillic INSERT values)

Client A: ProcessExtendedBuffer
  → reads Parameters → sees Client B's data → sends garbage to shard
```

## Scope of affected fields

Only `[][]byte` fields decoded via sub-slicing need deep-copy. Other fields are safe:

| Field | Type | Safe? | Why |
|-------|------|-------|-----|
| `Bind.DestinationPortal` | `string` | Yes | `string()` copies bytes |
| `Bind.PreparedStatement` | `string` | Yes | `string()` copies bytes |
| `Bind.ParameterFormatCodes` | `[]int16` | Yes | `make([]int16, ...)` with `Uint16` reads |
| `Bind.ResultFormatCodes` | `[]int16` | Yes | `make([]int16, ...)` with `Uint16` reads |
| **`Bind.Parameters`** | `[][]byte` | **No** | **sub-slices of src** |
| `Parse.Name`, `Parse.Query` | `string` | Yes | `string()` copies bytes |
| `Parse.ParameterOIDs` | `[]uint32` | Yes | read via `buf.Next(4)` copy |
| **`FunctionCall.Arguments`** | `[][]byte` | **No** | **sub-slices of src** |
| **`DataRow.Values`** | `[][]byte` | **No** | **sub-slices of src** |

## Fix

Deep-copy all retained `[][]byte` fields at the earliest retention point:

- **`Bind.Parameters`** in `frontend.ProcessMessage` — primary fix for the reported bug
- **`FunctionCall.Arguments`** in `frontend.ProcessMessage` — same sub-slicing pattern
- **`DataRow.Values`** in `proxy_routing.go` — two locations where values are stored across shard `Receive()` iterations (`colValues` in distributed INSERT, `tts.Raw` in scatter plan)

A shared `xproto.CopyByteSlices` helper is added, following the existing `append([]byte(nil), ...)` idiom already used in `relay/executor.go` for `cacheCC.CommandTag`.

## Files changed

- **New**: `router/xproto/copy.go` — `CopyByteSlices([][]byte) [][]byte` helper
- **Edit**: `router/frontend/frontend.go` — deep-copy `Bind.Parameters` and `FunctionCall.Arguments`
- **Edit**: `router/qrouter/proxy_routing.go` — deep-copy `DataRow.Values` in two locations

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./router/...` passes (all existing tests)
- [x] `go test ./pkg/...` passes (all existing tests)
- [x] Deployed to production — confirmed the cross-contamination no longer occurs under the same concurrent workload that previously triggered it

### DIFF
--- a/router/frontend/frontend.go
+++ b/router/frontend/frontend.go
@@ -3,6 +3,7 @@ package frontend
 import (
 	"context"
 	"io"
+	"slices"
 	"time"
 
 	"github.com/jackc/pgx/v5/pgproto3"
@@ -16,6 +17,7 @@ import (
 	"github.com/pg-sharding/spqr/router/qrouter"
 	"github.com/pg-sharding/spqr/router/relay"
 	"github.com/pg-sharding/spqr/router/statistics"
+	"github.com/pg-sharding/spqr/router/xproto"
 )
 
 // ProcessMessage: process client iteration, until next transaction status idle
@@ -54,6 +56,7 @@ func ProcessMessage(qr qrouter.QueryRouter, rst relay.RelayStateMgr, msg pgproto
 		// copy interface
 		cpQ := *q
 		q = &cpQ
+		q.ParameterOIDs = slices.Clone(q.ParameterOIDs)
 
 		rst.AddExtendedProtocMessage(q)
 		return nil
@@ -68,6 +71,8 @@ func ProcessMessage(qr qrouter.QueryRouter, rst relay.RelayStateMgr, msg pgproto
 		// copy interface
 		cpQ := *q
 		q = &cpQ
+		q.Arguments = xproto.CopyByteSlices(q.Arguments)
+
 		spqrlog.Zero.Debug().
 			Uint("client", rst.Client().ID()).
 			Msg("client function call: simply fire parse stmt to connection")
@@ -85,6 +90,9 @@ func ProcessMessage(qr qrouter.QueryRouter, rst relay.RelayStateMgr, msg pgproto
 		// copy interface
 		cpQ := *q
 		q = &cpQ
+		q.Parameters = xproto.CopyByteSlices(q.Parameters)
+		q.ResultFormatCodes = slices.Clone(q.ResultFormatCodes)
+		q.ParameterFormatCodes = slices.Clone(q.ParameterFormatCodes)
 
 		rst.AddExtendedProtocMessage(q)
 		return nil
--- a/router/qrouter/proxy_routing.go
+++ b/router/qrouter/proxy_routing.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pg-sharding/spqr/router/rmeta"
 	"github.com/pg-sharding/spqr/router/server"
 	"github.com/pg-sharding/spqr/router/virtual"
+	"github.com/pg-sharding/spqr/router/xproto"
 
 	"github.com/pg-sharding/lyx/lyx"
 )
@@ -473,7 +474,7 @@ func (qr *ProxyQrouter) planQueryV1(
 						for _, is := range iis {
 							for _, col := range is.Columns {
 								colValues[col] = append(colValues[col],
-									v.Values[ind])
+									slices.Clone(v.Values[ind]))
 								ind++
 							}
 						}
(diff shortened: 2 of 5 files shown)


reshke:
Yes, bugs of this kind are well-known.
We can also use CopyByteSlices inside `gangMemberDeployPreparedStatement`  pgproto3.RowDescription handling

Ledatu:
I've introduced `CopyFieldDescriptions` in the `xproto` package

Since `RowDescription.Fields` is a slice of structs, I used the same deep-copy pattern as in `CopyByteSlices` to ensure the `Name` fields are safely cloned. This keeps the implementation in both `gangMemberDeployPreparedStatement` and `sliceDescribePortal` clean and consistent

I also added unit tests that verify mutation isolation and nil-safety

Regarding CI: the `regress_ldap_server_2` failure (LDAP exit 255) looks unrelated to these changes

Denchick (dismissed):
I like this implementation for simplicity. But seems like you should to fix tests.

Ledatu:
I'll try to do it today, seems like all my PR's don't want to pass the tests now :)

Ledatu:
Merged upstream for USE_PGXS fix

reshke:
I have revisited this PR and pushed new version. I used more idiomatic slice.Clone in few places, and adjusted unit tests. I plan to commit this today.

Issue #1764, opened 2025-11-19
Check shard connectivity on AddShard


Denchick:
Problem: you may add now shards that is not reachably by coordinator or router:

```
regress=> CREATE SHARD sh5 WITH HOSTS oooops:6432;
    add shard    
-----------------
 shard id -> sh5
(1 row)
```

For example, on the router startup we check can connect to each shard (no proofs)

Daewoow:
Hello! I'd like to work on this issue

PR #2178, opened 2026-03-14
Fix: Issue #1764 - reject unreachable shard hosts on CREATE SHARD and AddDataShard 

Daewoow:
This PR addresses a bug where a shard could be added with an unreachable host (for example, oooops:6432) without immediate failure.
Before this change, CREATE SHARD ... WITH HOSTS ... could succeed and persist metadata, while the actual error only appeared later on the first real backend connection attempt.

Now, unreachable shard hosts are rejected immediately.

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -2,6 +2,7 @@ package meta
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"syscall"
@@ -585,6 +586,23 @@ func ProcessCreate(ctx context.Context, astmt spqrparser.Statement, mngr EntityM
 		}
 		return tts, nil
 	case *spqrparser.ShardDefinition:
+		_, err := mngr.GetShard(ctx, stmt.Id)
+		if err == nil {
+			return nil, spqrerror.Newf(spqrerror.SPQR_INVALID_REQUEST, "shard with id %s already exists", stmt.Id)
+		}
+
+		var spErr *spqrerror.SpqrError
+		if errors.As(err, &spErr) {
+			if spErr.ErrorCode != spqrerror.SPQR_NO_DATASHARD {
+				return nil, err
+			}
+		} else {
+			cleanErr := spqrerror.CleanGrpcError(err)
+			if !strings.Contains(cleanErr.Error(), "unknown shard") {
+				return nil, cleanErr
+			}
+		}
+
 		shardCfg := &config.Shard{
 			RawHosts: stmt.Hosts,
 			Type:     config.DataShard,
@@ -598,7 +616,11 @@ func ProcessCreate(ctx context.Context, astmt spqrparser.Statement, mngr EntityM
 				RootCertFile: stmt.RootCertFile,
 			}
 		}
+
 		dataShard := topology.NewDataShard(stmt.Id, shardCfg)
+		if err := topology.ValidateDataShardHosts(ctx, dataShard); err != nil {
+			return nil, err
+		}
 		if err := mngr.AddDataShard(ctx, dataShard); err != nil {
 			return nil, err
 		}
--- a/pkg/models/topology/validation.go
+++ b/pkg/models/topology/validation.go
@@ -1,0 +1,60 @@
+package topology
+
+import (
+	"context"
+	"net"
+	"sync"
+	"time"
+
+	"github.com/pg-sharding/spqr/pkg/models/spqrerror"
+	"github.com/pg-sharding/spqr/pkg/spqrlog"
+)
+
+const defaultShardHostValidationTimeout = 3 * time.Second
+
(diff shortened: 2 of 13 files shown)


test/feature/features/spqrguard.feature
```diff
@@ -23,6 +23,7 @@ Scenario: spqrguard is set up correctly
     CREATE TABLE t(id int)
     """
     Then command return code should be "0"
+    And we wait for "5" seconds
```
Denchick: ahaha, why?

pkg/models/topology/validation.go
```diff
@@ -0,0 +1,41 @@
...
+	hosts := shard.Cfg.Hosts()
+	if len(shard.Cfg.RawHosts) != 0 && len(hosts) != len(shard.Cfg.RawHosts) {
+		return spqrerror.Newf(spqrerror.SPQR_INVALID_REQUEST, "shard %q has invalid or unsupported host definitions", shard.ID)
+	}
+	if len(hosts) == 0 {
+		return spqrerror.Newf(spqrerror.SPQR_INVALID_REQUEST, "shard %q has no valid hosts configured", shard.ID)
+	}
+
+	dialer := &net.Dialer{Timeout: defaultShardHostValidationTimeout}
+	for _, host := range hosts {
```
Denchick, line 35: 3-second timeout per host is sequential. If a shard has N hosts and multiple are unreachable, the worst case is N * 3s of blocking.

Consider to validate shard hosts in parallel.

pkg/meta/meta.go
```diff
@@ -582,10 +582,19 @@ func ProcessCreate(ctx context.Context, astmt spqrparser.Statement, mngr EntityM
 		}
 		return tts, nil
 	case *spqrparser.ShardDefinition:
+		if _, err := mngr.GetShard(ctx, stmt.Id); err == nil {
+			return nil, spqrerror.Newf(spqrerror.SPQR_INVALID_REQUEST, "shard with id %s already exists", stmt.Id)
+		} else if !strings.Contains(err.Error(), "unknown shard") {
```
Denchick: If that string ever changes, this check silently breaks. Better to use a typed error or error code comparison

pkg/models/topology/validation.go
```diff
@@ -0,0 +1,41 @@
...
+		conn, err := dialer.DialContext(ctx, "tcp", host)
+		if err != nil {
+			return spqrerror.Newf(spqrerror.SPQR_INVALID_REQUEST, "shard host %q is not reachable: %v", host, err)
+		}
+		if err := conn.Close(); err != nil {
+			return spqrerror.Newf(spqrerror.SPQR_INVALID_REQUEST, "failed to close validation connection to shard host %q: %v", host, err)
```
Denchick: Failing to close a validation socket should not prevent shard creation.
You may log it as a warning

Denchick:
Thank you for your PR! I will review it soon

pkg/config/router.go
```diff
@@ -186,7 +187,16 @@ func (s *Shard) parseHosts() {
 			host.AZ = parts[2]
 			host.Address = fmt.Sprintf("%s:%s", parts[0], parts[1])
 		} else {
-			host.Address = rawHost
+			addr := rawHost
```
Denchick: i don't understand why do you need these changes?

test/feature/features/spqrguard.feature
Daewoow: The tests didn't pass, and I thought this would help =). It turns out, it didn't make any difference

pkg/models/topology/validation.go
Daewoow: I fixed it.

pkg/meta/meta.go
Daewoow: I changed it

pkg/models/topology/validation.go
Daewoow: I corrected it, thanks.

pkg/config/router.go
Daewoow: At some point, I thought that the tests were failing because of this. Rolled back the change

pkg/meta/meta.go
```diff
@@ -584,10 +585,29 @@ func ProcessCreate(ctx context.Context, astmt spqrparser.Statement, mngr EntityM
 		}
 		return tts, nil
 	case *spqrparser.ShardDefinition:
+		if _, err := mngr.GetShard(ctx, stmt.Id); err == nil {
```
Denchick:
```
if smth {
    return xxx
} else {
    retrun yyy
}
```

is absolutely the same as 

```
if smth {
    return xxx
} 
retrun yyy
```

Daewoow: I'll fix it, thanks

PR #2186, opened 2026-03-15
Support renaming relation distribution keys

### LINKED ISSUE #1846: ALTER DISTRIBUTION do smth with key range
**Is your feature request related to a problem? Please describe.**
There is no way to rename a sharding key

TODO https://t.me/c/1592474610/1396

Ledatu:
## Summary

- add support for `ALTER DISTRIBUTION ... ALTER RELATION ... RENAME DISTRIBUTION KEY`
- rename the target distribution key column in relation metadata in place
- add parser, metadata, and console regression coverage for the new command

## Motivation

Needed this to rename a distribution key in production without detaching and re-attaching the relation or editing metadata manually

## Test plan

- `go test ./yacc/console`
- `go test ./pkg/meta`
- console regression coverage in `test/regress/tests/console/alter_relation`

Closes #1846

### DIFF
--- a/test/regress/tests/console/expected/alter_relation.out
+++ b/test/regress/tests/console/expected/alter_relation.out
@@ -45,9 +45,9 @@ ALTER DISTRIBUTION ds1 ALTER RELATION a RENAME DISTRIBUTION COLUMN another TO re
 (3 rows)
 
 SHOW relations;
- relation_name | distribution_id |     distribution_key      | schema_name  
----------------+-----------------+---------------------------+--------------
- a             | ds1             | ("renamed_col", identity) | $search_path
+ relation_name | distribution_id |     distribution_key      | schema_name 
+---------------+-----------------+---------------------------+-------------
+ a             | ds1             | ("renamed_col", identity) | public
 (1 row)
 
 ALTER DISTRIBUTION ds1 ALTER RELATION a RENAME DISTRIBUTION COLUMN nonexistent TO fail_col;


Denchick:
This PR seems related to https://github.com/pg-sharding/spqr/issues/1846

Thanks! I will try to review it soon

Ledatu:
I think it should work now - after today's spqrhash update, the Dockerfile on my branch wasn't updated to use USE_PGXS=1. I'll merge upstream into my other PRs too. Looking forward to your review!

reshke:
Its not `ALTER DISTRIBUTION ALTER RELATION RENAME DISTRIBUTION KEY` but `ALTER DISTRIBUTION ALTER RELATION RENAME DISTRIBUTION COLUMN` ?  We can have multi-column distribution key. Also, we need documentation for this, and we need to doc that this command does not have well-defined behaviour in case of concurrent read-write ops. Like, what should user do to migrate from one distribution scheme to another, on-line? How do we check that old and new distribution keys splits data in the same shards? So, this is setup and test only for now.

test/regress/tests/console/sql/alter_relation.sql
```diff
@@ -7,6 +7,11 @@ SHOW relations;
 ALTER DISTRIBUTION ds1 ALTER RELATION a DISTRIBUTION KEY another;
 SHOW relations;
 
+ALTER DISTRIBUTION ds1 ALTER RELATION a RENAME DISTRIBUTION KEY another TO renamed_col;
+SHOW relations;
+
+ALTER DISTRIBUTION ds1 ALTER RELATION a RENAME DISTRIBUTION KEY nonexistent TO fail_col;
```
reshke: SHOW relations; after that to ensure no changes done.

Ledatu:
Renamed to `RENAME DISTRIBUTION COLUMN`, added `SHOW relations;` after the failing rename, and added docs with concurrency warnings. Also documented the previously undocumented `ALTER RELATION DISTRIBUTION KEY` and `ALTER RELATION SCHEMA` commands

Note: The original issue #1846 could technically be solved with `ALTER RELATION DISTRIBUTION KEY`, but the new `RENAME DISTRIBUTION COLUMN` provides proper validation and saves the user from having to re-enter hash functions and other columns of composite keys if they made a typo in one of them, making it a worthwhile addition anyway

pkg/meta/meta.go
```diff
@@ -851,6 +851,50 @@ func processAlterRelation(ctx context.Context, astmt spqrparser.Statement, mngr
 			},
 		}
 
+		return tts, nil
+	case *spqrparser.RenameDistributionColumn:
+		ds, err := mngr.GetDistribution(ctx, dsId)
+		if err != nil {
+			return nil, err
+		}
+		rel := ds.Relations[relationName.RelationName]
```
reshke: ds.GetRelation()

pkg/meta/meta.go
```diff
@@ -851,6 +851,50 @@ func processAlterRelation(ctx context.Context, astmt spqrparser.Statement, mngr
...
+		if rel == nil {
+			return nil, spqrerror.Newf(spqrerror.SPQR_INVALID_REQUEST,
+				"relation \"%s\" is not attached to distribution \"%s\"", relationName.String(), dsId)
+		}
+		newKey := make([]distributions.DistributionKeyEntry, len(rel.DistributionKey))
+		copy(newKey, rel.DistributionKey)
+		found := false
+		for i, entry := range newKey {
```
reshke: Looks like we can use some utility function here. I would prefer defined func (rel * DistributedRelation) Something()

Also, in case of expression routing, entry.Column would be "".

reshke:
Try to run compose and run expression routing + this feature. I suspect this would fail.
https://github.com/pg-sharding/spqr/blob/master/test/regress/tests/router/sql/expression_routing.sql#L4

pkg/models/distributions/distribution.go
```diff
@@ -38,6 +38,60 @@ func (r *DistributedRelation) QualifiedName() rfqn.RelationFQN {
 	return *r.Relation
 }
 
+// RenameKeyColumn returns a copy of the distribution key with oldName replaced by newName.
+// Rejects the rename if the column is referenced by a unique index or if the
+// relation uses expression-based routing only.
+func (r *DistributedRelation) RenameKeyColumn(oldName, newName string) ([]DistributionKeyEntry, error) {
+	hasColumnEntries := false
+	for _, entry := range r.DistributionKey {
```
reshke: Lets define IsExpressionRouting() & GetColumn() methods of DistributedRelation and use them - they can be reused in number of places

test/regress/tests/router/sql/rename_column_routing.sql
```diff
@@ -0,0 +1,46 @@
...
+SELECT * FROM rc_test WHERE w_id = 20;
+
+DROP TABLE rc_test;
+
+-- Expression routing: rename must be rejected
+\c spqr-console
+
+CREATE DISTRIBUTION ds_expr COLUMN TYPES int hash;
+ALTER DISTRIBUTION ds_expr ATTACH RELATION expr_test DISTRIBUTION KEY MURMUR [id1 INT HASH, id2 VARCHAR HASH];
+ALTER DISTRIBUTION ds_expr ALTER RELATION expr_test RENAME DISTRIBUTION COLUMN id1 TO new_id;
```
reshke: Isn't this checked in alter_relation.sql of console regression suite?

Ledatu: You are right, I'll remove it, main goal of this test is the routing-before-and-after scenario, is this justified to be tested out?

reshke: This 3 lines just ensure that router respond with correct errcode/errmessage, so we only need to check this once, and console-regress (alter_relation.sql) is just a better place

reshke:
I m looking forward to committing this. Please rebase this on master and run `make generate`.
I'll make a quick pass after this, maybe toggle some wording or kibitzing on other things, and commit.

PR #2228, opened 2026-03-25
executor.go: pass parameter status to client

rkhapov:
Vanilla pg do that


reshke:
wel this doesnt do...

PR #2231, opened 2026-03-25
Store two phase tx metadata storage shards in etcd

EinKrebs:


test/regress/tests/console/expected/show_two_phase_tx_metadata_storage.out
reshke, line 1: Looks like this regression test is missing from console schedule

reshke (approved):

PR #2270, opened 2026-04-06
Covert few more test targets to user generic test runner

reshke:


Denchick:
Thanks for the PR. I’m going to close this one because it looks stale now and no longer applies cleanly to the current master.

PR #2298, opened 2026-04-08
Enable role system in regression files

reshke:


test/regress/conf/router.yaml
```diff
@@ -90,3 +90,8 @@ schema_cache_backend_rule:
   db: regress
   pool_discard: true
   pool_rollback: true
+
+
+enable_role_system: true
```
Denchick, line 95: are you kidding https://github.com/pg-sharding/spqr/blob/f6b012c67d8695164a10f49c082e118a45ce52e8/test/regress/conf/router.yaml#L14

Denchick:
Please do not merge it as is. It is not acceptable.

PR #2351, opened 2026-04-19
feat(cmd/coordinator): add 'run' subcommand for CLI consistency

### LINKED ISSUE #1715: Unify CLI command style between `spqr-router` and `spqr-coordinator`
Currently, the `spqr-router` binary uses a run subcommand to start, while `spqr-coordinator` starts directly without it.
This inconsistency can be confusing for users switching between the two components.

Examples:

```
# Router requires 'run'
./spqr-router run -c examples/localrouter.yaml

# Coordinator starts without 'run'
./spqr-coordinator -c examples/coordinator.yaml

# Coordinator rejects 'run'
./spqr-coordinator run -c examples/coordinator.yaml
# -> Error: unknown command "run" for "spqr-coordinator"
```

**Expected behavior**:
Both binaries should follow the same CLI pattern — start with `run` command. The coordinator binary should support backward compatibility -- start without `run` as well.

tmchow:
## Summary

Unifies CLI style between `spqr-router` and `spqr-coordinator` per #1715. The router already uses `spqr-router run -c path-to-config`, while the coordinator starts directly without `run` and previously errored on `spqr-coordinator run -c ...`.

## Changes

- Extract the coordinator startup logic out of `rootCmd.RunE` into a shared `runCoordinator(cmd *cobra.Command) error` function.
- Add a `run` subcommand (`runCmd`) that calls `runCoordinator`.
- Keep `rootCmd.RunE` invoking `runCoordinator` too, so `spqr-coordinator -c path-to-config` (no subcommand) still works as before.

## Behavior

```
$ spqr-coordinator run -c coordinator.yaml    # new, matches spqr-router
$ spqr-coordinator -c coordinator.yaml        # still works (backward compat)
$ spqr-coordinator test-config ...            # unchanged
```

`spqr-coordinator --help` now also lists `run` among available commands.

## Verification

- `go vet ./cmd/coordinator/...` passes
- `go build -o /tmp/spqr-coord-build ./cmd/coordinator` succeeds
- `go test ./cmd/coordinator/...` passes (10 tests)

Fixes #1715

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/cmd/coordinator/main.go
+++ b/cmd/coordinator/main.go
@@ -42,6 +42,83 @@ func getMaxTxnBatchSize(configCoord *config.Coordinator) uint16 {
 	return maxTxnBatchSize
 }
 
+func runCoordinator(cmd *cobra.Command) error {
+	cfgStr, err := config.LoadCoordinatorCfg(cfgPath)
+	if err != nil {
+		return err
+	}
+	log.Println("Running config:", cfgStr)
+
+	if config.CoordinatorConfig().EnableRoleSystem {
+		if config.CoordinatorConfig().RolesFile == "" {
+			return fmt.Errorf("role system enabled but no roles file specified, see `enable_role_system` and `roles_file` in config")
+		}
+		rolesCfgStr, err := config.LoadRolesCfg(config.CoordinatorConfig().RolesFile)
+		if err != nil {
+			return err
+		}
+		log.Println("Running roles config:", rolesCfgStr)
+	}
+
+	if logLevel != "" {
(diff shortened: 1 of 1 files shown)


Denchick:
Thanks for your contribution.

However, this issue is intentionally designed as an onboarding task for new human contributors to get familiar with the project and contribution process. It does not provide meaningful value to the codebase itself.

We do not accept PRs like this, as they defeat the purpose of the issue.

PR #2391, opened 2026-04-26
protoTestRunner: check for extra messages

rkhapov:


reshke (approved):

rkhapov:
need to find a better way to do that, maybe just run the test several times

PR #2410, opened 2026-04-28
Fix converting generic option action from SQL

diPhantxm:


reshke (commented):
Any test to exercise this, by chance?

diPhantxm:
> Any test to exercise this, by chance?

I extended ALTER query in regress test with explicit ADD keywork, this should fail without fix. Because prior to this fix, we converted from one enum to another without any mapping, and the order is different. So in SQL parser ADD has value 2, and in our logic ADD has value 1. That resulted in ADD behaved unexpectedly as SET

reshke (approved):

PR #2419, opened 2026-04-29
Add automatic router re-bootstrap to coordinator

EinKrebs:


EinKrebs:
Close in favour of #2957

PR #2440, opened 2026-05-01
Implement LOCAL & GLOBAL show kinds

reshke:


yacc/console/ast.go
```diff
@@ -418,6 +420,15 @@ type InstanceControlPoint struct {
 
 func (*InstanceControlPoint) iStatement() {}
 
+//revive:disable:var-naming
```
Denchick, line 423: Don't overdo it. If there is an opportunity and a desire, it is better to avoid nolint.

Denchick (approved):

PR #2455, opened 2026-05-03
qdb meta transaction infra for AlterSequenceAttach

xelavopelk:
qdb meta transaction infra for AlterSequenceAttach
issue #1737


Denchick:
We've decided to proceed with the key range versioning approach instead, which makes this change unnecessary. Closing this PR in favor of that direction.

PR #2471, opened 2026-05-04
Fix govet inline warnings for generic slices calls

Denchick:
`govet` is a set of analyzers built into Go. One of them — `inline` — checks whether the Go compiler can inline a function call, meaning substitute the function body directly at the call site instead of making an actual function call. This is an optimization that eliminates the overhead of a function call.

The problem was that `slices.ContainsFunc` and `slices.Delete` are generic functions (with type parameters). The Go compiler doesn't yet know how to inline generic calls when types are inferred automatically (type inference). The `inline` analyzer in the newer version of golangci-lint detects this and reports an error.

Plain code without generics is inlined by the compiler without issues, so the analyzer no longer complains.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,6 @@ require (
 	go.etcd.io/etcd/client/v3 v3.6.11
 	go.uber.org/mock v0.6.0
 	golang.org/x/crypto v0.50.0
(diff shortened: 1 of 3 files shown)


Denchick:
Closed in prior to https://github.com/pg-sharding/spqr/pull/2473

PR #2472, opened 2026-05-04
Remove obsolete stress tests

Denchick:
`test/stress/` are obsolete and no longer serve a useful purpose:

- **Replaced by proper benchmarks.** The project now has a dedicated benchmarking infrastructure (`benchmarks/`, `.github/workflows/benchmark.yaml`) that runs industry-standard TPC-C workloads via [benchbase](https://github.com/dvkashapov/benchbase-spqr) on production-grade hardware in Yandex Cloud (8-core shards, 32-core router, 32 GB RAM each). Benchmarks are triggered on any PR by adding the `performance` label and results are posted as a comment with a link to a full report.
- **The old stress tests were not meaningful.** They ran basic `pgbench select-only` and a simple deadlock test inside lightweight Docker containers with no resource controls — not representative of real-world performance characteristics. The custom `stress.go` tool was equally simplistic (random single-row SELECTs with hardcoded parameters).
- **They slowed down CI.** The `stress` job was a required dependency (`needs`) for `regress`, `regress-clustered`, `isolation-regress`, and `feature_prepare` jobs, adding unnecessary serialization to the test pipeline.


Denchick:
In test/stress/f.sql there is smth important, so you can't just drop stress tests. Close it

PR #2551, opened 2026-05-19
fix tests flap: isolate auth fixtures from feature tests

Denchick:
Move LDAP/KDC feature-test services behind an auth Compose profile so non-auth scenarios no longer fail on auth fixture startup. Auth router configs enable the profile automatically, and the test composer now waits for healthy containers before running steps.

This PR fixes https://github.com/pg-sharding/spqr/actions/runs/26088605906/job/76711202784?pr=2548


Denchick:
More cleaner version https://github.com/pg-sharding/spqr/pull/2573

PR #2556, opened 2026-05-20
Fix/out of range binds panic

Ledatu:
## Summary

- Validate prepared statement bind parameter format-code counts before routing.
- Return a protocol violation for malformed bind metadata instead of allowing router-side parameter resolution to panic.
- Add bounds checks and focused tests for missing or out-of-range bind parameters.

## Motivation

Before this, the bind handler just blindly trusted that the parameter format-code metadata would always match up with the actual payload. If a client sent garbage or inconsistent bind data, the routing code would try to index past the available parameters while resolving values, blowing things up. In my case caused the router to restart on oversized queries, making normal clients suffer too

This fix enforces that contract right at the protocol boundary and keeps the downstream routing paths safe from invalid index errors


pkg/prepstatement/prepstatement.go
```diff
@@ -37,13 +38,18 @@ type PreparedStatementMapper interface {
...
-func GetParams(paramsFormatCodes []int16, bindParams [][]byte) []int16 {
+func GetParams(paramsFormatCodes []int16, bindParams [][]byte) ([]int16, error) {
 	var queryParamsFormatCodes []int16
 	paramsLen := len(bindParams)
 
 	/* https://github.com/postgres/postgres/blob/c65bc2e1d14a2d4daed7c1921ac518f2c5ac3d17/src/backend/tcop/pquery.c#L664-L691 */ /* #no-spell-check-line */
 	if len(paramsFormatCodes) > 1 {
-		queryParamsFormatCodes = paramsFormatCodes
+		if len(paramsFormatCodes) != paramsLen {
+			return nil, spqrerror.Newf(spqrerror.PG_ERRCODE_PROTOCOL_VIOLATION,
```
reshke, line 48: It better to exercise this code with xproto test rather that unit test. One of immediate benefits is that yandex/odyssey will also run this test

something like 

https://github.com/pg-sharding/spqr/blob/master/test/xproto/xproto_test.go#L4159


the workflow is something like: create frontend message sequence, run it against raw posgtres, get the backend message sequence, then run it againts SPQR to verify your changes



test snippet

```
Request: []pgproto3.FrontendMessage{
				&pgproto3.Close{
					Name:       "corrupt_bind_num",
					ObjectType: 'S',
				},
				&pgproto3.Parse{
					Name: "p2corrupt_bind_num",
                     Query: "SELECT $1, $2"
				},
				&pgproto3.Bind{
					PreparedStatement: "p2",
					DestinationPortal: "",
/* no params, or 1 params */
				},
				&pgproto3.Sync{},
			},
```

Denchick:
@reshke review it please

reshke (approved):

PR #2568, opened 2026-05-25
fix: init_sql log last statement loss

### LINKED ISSUE #986: Wrong init_sql parsing
**Describe the bug**

The local log [iterates](https://github.com/pg-sharding/spqr/blob/master/router/qlog/provider/local.go#L63) over the file line by line, but it should iterate statement by statement.

**To Reproduce**

```
CREATE KEY RANGE uid_range_1
FROM 00000000-0000-0000-0000-000000000000
ROUTE TO shard1
FOR DISTRIBUTION uid_ds;
```

In the logs, you may see:

```
spqr-1    | {"level":"info","query":"CREATE KEY RANGE uid_range_1","time":"2025-01-30T08:30:11Z"}
spqr-1    | {"level":"error","error":"syntax error","time":"2025-01-30T08:30:11Z"}
spqr-1    | {"level":"error","error":"syntax error","time":"2025-01-30T08:30:11Z"}
spqr-1    | {"level":"info","query":"FROM 00000000-0000-0000-0000-000000000000","time":"2025-01-30T08:30:11Z"}
…
spqr-1    | {"level":"info","count":26,"filename":"/spqr-config/init.sql","time":"2025-01-30T08:30:11Z","message":"successfully init queries from 
file"}
```

**Expected behavior**
We should not fail on an incorrect init_sql format file, but we need to parse it by statements, not line endings.

NeSawer:
Fixes #986

The issue was partially resolved in #1011, but there was still a bug where the last statement would be lost if it didn't end with `;`.

In this PR, I'm fixing the remaining bug. It looks like the issue in #986 will be fully resolved.

### DIFF
--- a/router/qlog/provider/local.go
+++ b/router/qlog/provider/local.go
@@ -81,6 +81,10 @@ func (dw *LocalQlog) Recover(_ context.Context, path string) ([]string, error) {
 		}
 	}
 
+	if strings.TrimSpace(qs) != "" {
+		queries = append(queries, qs)
+	}
+
 	if err := scanner.Err(); err != nil {
 		return nil, err
 	}
--- a/router/qlog/provider/local.go
+++ b/router/qlog/provider/local.go
@@ -39,7 +39,6 @@ func (dw *LocalQlog) DumpQuery(_ context.Context, fname string, q string) error 
 	return nil
 }
 
-// TODO : unit tests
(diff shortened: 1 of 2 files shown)


Denchick (commented):
Solid work! Could you add a simple unit test for this?

reshke (approved):

PR #2576, opened 2026-05-26
fix: preserve routing metadata for derived table joins

Ledatu:
## Summary

- Preserve routing metadata for simple `FROM (SELECT ...) alias` derived tables by resolving alias columns back to passthrough columns from the underlying relation.
- Keep derived-table analysis scoped so inner table aliases and CTEs do not leak into the outer query, while still merging routing facts needed by the parent join.
- Add regression coverage for derived-table `LEFT JOIN` routing parity with the equivalent CTE shape, including passthrough `SELECT *` and renamed shard-key projection cases

## Motivation

An inline derived table on the preserved side of a `LEFT JOIN` could route differently from an equivalent CTE query. **In practice, this can silently dispatch to the wrong shard and return zero rows**, even when the left-side subquery has matching data

This change brings routing parity to simple derived-table aliases without broadening the router's scope to support arbitrary, complex derived-table semantics

### DIFF
--- a/router/planner/analyze.go
+++ b/router/planner/analyze.go
@@ -31,6 +31,124 @@ func analyzeFromClauseList(
 	return nil
 }
 
+func mergeRelationMap(dst, src map[rfqn.RelationFQN]struct{}) {
+	for rel := range src {
+		dst[rel] = struct{}{}
+	}
+}
+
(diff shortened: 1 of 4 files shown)


reshke:
I'll take a look and give feedback here soon.
Basically, Im +1 on everything that support what user needs, but I very unsure of this implementation

PR #2587, opened 2026-05-30
Fix log severity

diPhantxm:


reshke (approved):
Well, thats kind of temporal error, maybe WARN fits better here
anyway +1

PR #2588, opened 2026-06-02
Support UUID HASH column type in distributions

### LINKED ISSUE #2490: Support `UUID HASH` column type in distributions
```sql
spqr-console=> CREATE DISTRIBUTION main_event_uuid_distribution COLUMN TYPES uuid hash;
ERROR:  failed to parse query "CREATE DISTRIBUTION main_event_uuid_distribution COLUMN TYPES uuid hash;": syntax error
```

Currently, `UUID` is only supported as an identity (non-hashed) column type. The grammar (`gram.y:1497-1498`) recognizes `UUID` but has no `UUID HASH` variant, unlike `VARCHAR`/`VARCHAR HASH` and `INTEGER`/`INT HASH`. The distribution validation in `distribution.go:735-767` also explicitly forbids hash functions on UUID columns, and the hash function implementations (`ApplyMurmurHashFunction`, `ApplyCityHashFunction`) don't handle UUID input.

## Motivation

UUIDv4 values are already uniformly random, so identity-based distribution works well — no hashing is needed. However, UUIDv7 values are time-ordered (the first 48 bits encode a timestamp), which means identity-based distribution leads to hot shards as new rows cluster in the same key range.

Hashing UUIDv7 values would distribute them uniformly across shards, similar to how `INTEGER HASH` / `VARCHAR HASH` work today.

## Consideration

We need to decide how to handle the distinction between UUIDv4 and UUIDv7:

- **Option A**: Add a single `UUID HASH` type that always hashes. Users with UUIDv4 can still use plain `UUID` (identity). Users with UUIDv7 use `UUID HASH`.
- **Option B**: Introduce explicit `UUIDV4` / `UUIDV7` types with different defaults.

Denchick:
Fixes https://github.com/pg-sharding/spqr/issues/2490


pkg/plan/util.go
```diff
@@ -23,6 +23,8 @@ func ParseResolveParamValue(paramCode int16, ind int, tp string, bindParams [][]
 		case qdb.ColumnTypeUUID:
 			val := string(bindParams[ind])
 			return val, nil
+		case qdb.ColumnTypeUUIDHashed:
+			return string(bindParams[ind]), nil
```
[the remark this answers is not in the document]
Denchick: https://github.com/pg-sharding/spqr/issues/2591

test/regress/tests/router/sql/hash_routing.sql
```diff
@@ -32,7 +32,37 @@ The quick brown fox jumps over the lazy dog
...
+CREATE TABLE xxhashuuid (id uuid);
+INSERT INTO xxhashuuid (id) VALUES ('018f4b8e-37f0-7cc4-b5f2-0f62d09ca662');
+INSERT INTO xxhashuuid (id) VALUES ('018F4B8E-37F0-7CC4-B5F2-0F62D09CA663');
+INSERT INTO xxhashuuid (id) VALUES ('00000000-0000-0000-0000-000000000000');
+INSERT INTO xxhashuuid (id) VALUES ('ffffffff-ffff-ffff-ffff-ffffffffffff');
+
+SELECT * FROM xxhashuuid ORDER BY id /* __spqr__execute_on: sh1 */;
+SELECT * FROM xxhashuuid ORDER BY id /* __spqr__execute_on: sh2 */;
+
+COPY xxhashuuid (id) FROM STDIN;
```
reshke, line 58: also select where uuid.= test and __spqr__ctid test

test/regress/tests/router/sql/schema_qual.sql
```diff
@@ -53,6 +53,7 @@ DROP SCHEMA sh1 CASCADE;
 DROP SCHEMA sh2 CASCADE;
 
 SET __spqr__maintain_params TO false;
+SET search_path TO 'public';
```
reshke, line 56: thats bad, rebase

reshke:
+- ok

reshke (approved):

PR #2595, opened 2026-06-04
Check shard migrations on `CREATE SHARD`

diPhantxm:


Denchick (changes requested):
Nice feature idea! The scope here is pretty large though — 37 files across proto, coordinator, grammar, and tests. It's tough to review all of this as a single unit. Can we split it up? E.g.:

1. Grammar/parser changes
2. Everything else

Smaller PRs = faster reviews and fewer hidden issues.

diPhantxm:
> Nice feature idea! The scope here is pretty large though — 37 files across proto, coordinator, grammar, and tests. It's tough to review all of this as a single unit. Can we split it up? E.g.:
> 
> 1. Grammar/parser changes
> 2. Everything else
> 
> Smaller PRs = faster reviews and fewer hidden issues.

#2635 - PR with yacc changes only

pkg/coord/coord.go
```diff
@@ -1481,3 +1489,119 @@ func (lc *Coordinator) SetTwoPhaseTxMetaStorage(ctx context.Context, storage []s
...
+	// Check reference relations exist on shard
+	for _, refRel := range refRelations {
+		shardColumns, err := listShardColumns(ctx, conn, refRel.RelationName)
+		if err != nil {
+			return err
+		}
+
+		if len(shardColumns) == 0 {
+			return spqrerror.Newf(spqrerror.SPQR_OBJECT_NOT_EXIST,
+				"shard is not ready\nHINT: relation %s.%s does not exist on shard",
```
Denchick: There is an option to specify hint via method

pkg/bootstrap/bootstrap.go
```diff
@@ -47,7 +47,7 @@ func EtcdReBootstrap(ctx context.Context, mngr meta.EntityMgr, qdbAddrs []string
 		}
 		/* Add shards from coordinator */
 		for _, sh := range shards {
-			if err := mngr.AddDataShard(ctx, topology.DataShardFromDB(sh)); err != nil {
+			if err := mngr.AddDataShard(ctx, topology.DataShardFromDB(sh), true); err != nil {
```
Denchick, line 50: maybe true maybe not?

test/regress/tests/router/expected/create_shard.out
```diff
@@ -0,0 +1,83 @@
+\c spqr-console
```
Denchick, line 1: we may rebootstrap router with force:false (is it correct) and still may fail this validation, which will lead us to an incident
I think it is a good idea to test such a scenario

pkg/bootstrap/bootstrap.go
diPhantxm: I think it should be `true` here since we don't want to double check migrations on shard. Coordinator did it previously. We just want to copy metadata from coordinator into router

test/regress/tests/router/expected/create_shard.out
diPhantxm: Added a test scenario into feature-tests since we need use_coordinator_init to run rebootstrap from sql console

Denchick (approved):

Denchick (approved):

PR #2596, opened 2026-06-05
Fix router postprocessing regression flake

Denchick:
Fixes same problem but in postprocessing https://github.com/pg-sharding/spqr/pull/2592


Denchick:
Addressed Copilot's workflow comments by removing `grep` from the unhealthy-container log pipelines. `awk` now handles both filtering and printing container IDs, so the command exits successfully when there are no unhealthy containers while still logging unhealthy ones when present.

Denchick:
Split the CI workflow logging change into a separate PR: #2597. This PR now contains only the router `postprocessing` regression fix.

reshke (approved):

Issue #1887, opened 2026-01-05
Documentation KEY RANGE concept


Denchick:
User feedback

> There are no manuals for working with SPLIT KEY RANGE / UNION KEY RANGE. Setting the boundaries is not obvious, there is no example of splitting not for the first range in the source code.

We should update this page https://docs.pg-sharding.tech/sharding/console/concepts#key-range, where we describe what key range is, why there is no left border, why you should create it with your own, etc.

Denchick:
Related issue: https://github.com/pg-sharding/spqr/issues/1754

Denchick:
I’ve spent a lot of time reflecting on user feedback and came to the conclusion that there’s significant room for improvement in how we work with key ranges.

What I’m proposing is a change aimed at reducing the cognitive load for our users when configuring SPQR.

## Main pain points

1. Unclear how to properly set bounds for range and hash distributions.
2. Order of key range creation is non-obvious (user must create KRs in reverse order, wtf)
3. Intersection validation forces users to think about implementation details
4. MOVE vs REDISTRIBUTE confusion. MOVE only updates metadata, doesn't migrate data. wat?

## Key Ranges UX Improvement Proposals

### 1. Automatic Key Range Distribution

**Problem:** Users must manually calculate bounds for each shard.

**Solution:** New `CREATE KEY RANGES AUTO` command

```sql
-- Instead of this:
CREATE KEY RANGE key_range2 FROM 2147483648 ROUTE TO shard2 FOR DISTRIBUTION answers;
CREATE KEY RANGE key_range1 FROM 0 ROUTE TO shard1 FOR DISTRIBUTION answers;

-- This:
CREATE KEY RANGES FOR DISTRIBUTION answers
  USING SHARDS shard1, shard2;

-- Or even simpler:
CREATE KEY RANGES FOR DISTRIBUTION answers USING ALL SHARDS;
```

The system would automatically:
- Determine hash type (MURMUR → uint32, 0..4294967295)
- Evenly split the range across shards
- Create key ranges in correct order

### 2. Adding New Shard with Auto-rebalancing

**Scenario:** Added shard3, want to rebalance.

```sql
-- Current painful flow:
-- 1. Understand current bounds
-- 2. Calculate new boundaries
-- 3. SPLIT necessary ranges
-- 4. REDISTRIBUTE data

-- Proposed:
REBALANCE DISTRIBUTION answers
  USING SHARDS shard1, shard2, shard3;

-- Or when adding a shard:
[...]

PR #2604, opened 2026-06-08
Introduce `CREATE KEY RANGES FOR DISTRIBUTION`

diPhantxm:
Support creating key ranges for distribution without specifying boundaries. It splits full key range into equal parts. 
Solves #1887 `Automatic Key Range Distribution`

Added support for specifying custom distribution key range with `CREATE KEY RANGES .... BETWEEN 0 AND 1000`. So there will be created key ranges splitting range [0, 1000] onto N shards

Need to figure out how to split range equally if distribution key is composite. I am not sure if it's a rare case or not, I suggest adding support for composite keys later


Denchick (changes requested):
Nice feature! Same ask as #2595 -- let's split grammar changes(ast, gram.y, lex, reserved keywords, parser tests) into a separate PR from the logic + regress tests.

Denchick:
Merged https://github.com/pg-sharding/spqr/pull/2639

pkg/meta/key_range_meta.go
```diff
@@ -167,12 +174,139 @@ func createKeyRange(ctx context.Context, mngr *TranEntityManager, stmt *spqrpars
...
+	if err := mngr.CommitTran(ctx); err != nil {
+		return nil, err
+	}
+
+	return createdKrs, nil
+}
+
+func splitEqualFullKeyRange(colTypes []string, shardsNumber int, customRange *spqrparser.CustomDistributionRange) ([][][]byte, error) {
+	if len(colTypes) > 1 {
+		return nil, spqrerror.New(spqrerror.SPQR_NOT_IMPLEMENTED, "cannot determine key ranges for composite keys").Hint("Use CREATE KEY RANGE...")
```
Denchick, line 282: Hint: specify them explicitly

Denchick (changes requested):
Great work, several notes:

1. I'd suggest making `USING ALL SHARDS` the default behavior, so the user doesn't have to specify it explicitly. For example, `CREATE KEY RANGES FOR DISTRIBUTION ds1;` should work the same as `CREATE KEY RANGES FOR DISTRIBUTION ds1 USING ALL SHARDS;` — it's the most common use case and reduces verbosity.

2. Where are the docs for the new syntax? New SQL statements should be documented so users can discover and use them.
3. I think it is a good idea to update Get Started Guide as well, maybe i will do it myself later.

Denchick (approved):

PR #2614, opened 2026-06-10
fix: decode binary-format UUID bind params in routing

### LINKED ISSUE #2591: UUID binary bind params may be parsed incorrectly
While reviewing UUID HASH, we noticed unrelated existing behavior in `ParseResolveParamValue`:

```go
case xproto.FormatCodeBinary:
case qdb.ColumnTypeUUID:
    return string(bindParams[ind]), nil
```

For PostgreSQL type-level binary format, `uuid` is 16 raw bytes:
- `UUID_LEN 16` in PostgreSQL source
- `uuid_recv` reads `pq_getmsgbytes(buffer, UUID_LEN)`
- `uuid_send` writes `pq_sendbytes(..., uuid->data, UUID_LEN)`

Refs:
- https://github.com/postgres/postgres/blob/master/src/include/utils/uuid.h
- https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/uuid.c

So `string(bindParams[ind])` may be wrong for real binary UUID params.

mvanhorn:
## Summary
Routing now decodes UUID bind parameters sent in binary format. PostgreSQL drivers that bind UUIDs with format code 1 send the raw 16-byte representation; `pkg/plan/util.go` previously only handled the text form, so binary-format UUID params failed to resolve to a routing key. The 16-byte value is now converted to the canonical hyphenated string, with the existing text path unchanged as the fallback.

## Why this matters
Issue #2591 reports that prepared statements with binary UUID bind params are not routed correctly, which affects any driver that defaults to binary format for the uuid OID (JDBC, npgsql, and Go's pgx in binary mode). The conversion mirrors how the surrounding code handles the other fixed-width binary types in the same switch.

## Testing
New table-driven tests in `pkg/plan/util_test.go` cover the 16-byte binary form (canonical string out), the existing text form, and malformed lengths falling through to the previous behavior. `gofmt`, `go vet`, and `go test ./pkg/plan/` pass.

Fixes #2591

### DIFF
--- a/pkg/plan/util.go
+++ b/pkg/plan/util.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/google/uuid"
 	"github.com/pg-sharding/spqr/qdb"
 	"github.com/pg-sharding/spqr/router/xproto"
 )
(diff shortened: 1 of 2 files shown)


Denchick:
Sorry, this issue is currently intended for human contributors only.

The implementation details are still under discussion in [pg-sharding/spqr#2614](https://github.com/pg-sharding/spqr/pull/2614), so we are not accepting automated or drive-by contributions for this one yet.

PR #2747, opened 2026-07-20
Security: Coordinator gRPC connection uses insecure transport (no TLS)

Denchick:


Denchick:
I think it is useful to separate the configuration types into server TLS and client TLS, because they have different semantics and validation. However, I would avoid treating `coordinator.yaml` as a shared TLS configuration file for both the coordinator and routers.

Both processes participate in the interconnect in both roles:

- the coordinator serves its gRPC API and also dials router gRPC APIs;
- the router serves its gRPC API and also dials the coordinator.

With mTLS, the coordinator server private key belongs only to the coordinator, while each router client private key belongs to that router. Putting both halves of a connection into a shared coordinator config either exposes credentials to processes that do not need them or encourages all routers to reuse one client identity. It also couples certificate rotation for independently deployed components. The balancer and CLI clients may need separate client identities as well.

There is also a current implementation detail to consider: `--coordinator-config` is always available as a flag, but the router only loads that file when `WithCoordinator`, `UseCoordinatorInit`, or `StoreTxDataPostgresql` is enabled. Therefore it is not currently an unconditional router configuration source.

My preference is to keep TLS configuration component-local and peer-qualified:

- `router.yaml`: router gRPC server TLS plus router→coordinator client TLS;
- `coordinator.yaml`: coordinator gRPC server TLS plus coordinator→router client TLS;
- `balancer.yaml`/CLI flags: their own coordinator client TLS.

The existing `grpc_api_tls`, `coordinator_grpc_tls`, and `router_grpc_tls` naming expresses that ownership and direction fairly well. If we want a more symmetric shape, something such as `grpc_tls.server` and `grpc_tls.clients.coordinator` / `grpc_tls.clients.routers` would remain clear. A generic pair of `grpc_server_tls` and `grpc_client_tls` is reasonable only inside each component's own config; it becomes ambiguous if those two fields are meant to describe the whole interconnect from one shared file.

We can still centralize the shared CA/trust profile through deployment tooling or a secret manager, without centralizing leaf private keys in `coordinator.yaml`.

PR #2760, opened 2026-07-23
Use SET LOCAL in flux util slice

reshke:


Denchick (approved):

Denchick:
Fixes
- #2696
- #2701

PR #2761, opened 2026-07-23
Add workflow to retry failed tests

diPhantxm:


Denchick (changes requested):
Image

Denchick:
Let's just write stable tests please guys

Denchick:
https://github.com/pg-sharding/spqr/pull/2756 +

diPhantxm:
> Let's just write stable tests please guys

well, true, but I anyway constantly retry tests because they are just flaky
Or we could re-use this workflow to retry some workflows which suffer from infrastructure fails

diPhantxm:
@Denchick 
Are you still sure we don't want this workflow? I have a PR in which I retried tests 3 times. Two of them were not because tests are flaky but because of network or something... The third retry was because of `postprocessing` regress test
https://github.com/pg-sharding/spqr/actions/runs/31703967595?pr=2887

Denchick:
We have to make our tests stable; adding an automatic retry feature is a slippery slope.

PR #2784, opened 2026-07-29
Fix task worker goroutine leaks

Denchick:


coordinator/pkg/clustered_coord.go
```diff
@@ -1385,14 +1385,15 @@ func (qc *ClusteredCoordinator) executeMoveInternal(
 
 	execCtx, cancel := context.WithCancel(ctx)
 
-	ch := make(chan error)
```
diPhantxm, line 1388: Maybe let's read out errors from another goroutine if `NOWAIT` was specified?
Not sure we want to lose an error without being logged

diPhantxm:
I would suggest something like this
It guarantees we update the status, drop the lock and log the error if running with `nowait`
What do you think?
```
func (qc *ClusteredCoordinator) executeMoveInternal(
	ctx context.Context,
	taskGroup *tasks.MoveTaskGroup,
	nowait bool,
	invalidateTG bool,
	icpCH icp.ICPContextHolder,
) error {

	/* TODO: do not lose move data goroutine here,
	* remember its id in structure similar to client pool. */

	if invalidateTG && taskGroup != nil {
		qc.invalidateTaskGroupCache(taskGroup.ID)
	}

	execCtx, cancel := context.WithCancel(ctx)

	ch := make(chan error, 1)
	qc.dataTransferWorkers.Store(taskGroup.ID, &transferworker.TaskGroupWorkerState{
		Cancel: cancel,
	})

	qc.moveTaskWatcherInit.Do(qc.bootstrapWatcher(context.TODO()))

	go func() {
		ch <- qc.executeMoveTaskGroup(execCtx, taskGroup, icpCH)
		qc.dataTransferWorkers.Delete(taskGroup.ID)
	}()

	awaitResult := func() error {
		select {
		case <-ctx.Done():
			return spqrerror.NewByCode(spqrerror.SPQR_TRANSFER_ERROR)
		case err := <-ch:
			if err != nil {
				_ = qc.db.DropTaskGroupLock(ctx, taskGroup.ID)
				_ = qc.QDB().WriteTaskGroupStatus(ctx, taskGroup.ID, &qdb.TaskGroupStatus{State: string(tasks.TaskGroupError), Message: err.Error()})
			}
			return err
		}
	}

	if nowait {
		go func() {
			if err := awaitResult(); err != nil {
				spqrlog.Zero.Error().Err(err).Str("task group id", taskGroup.ID).Msg("background move task group failed")
			}
		}()
		return nil
	}

	defer cancel()
	return awaitResult()
}
```

diPhantxm (approved):

PR #2824, opened 2026-08-03
Fix panic: nil conn on listener error

diPhantxm:


router/instance/instance.go
```diff
@@ -249,8 +249,10 @@ func (r *InstanceImpl) Run(ctx context.Context, listener net.Listener, pt port.R
 		for {
 			c, err := l.Accept()
 			if err != nil {
-				// handle error (and then for example indicate acceptor is down)
-				cChan <- nil
+				spqrlog.Zero.Error().
```
reshke, line 256: So, I get that passing nil to channel is bad, but are we sure that shutting down at single network error is good? 
Can we just continue and try to get new connection? Maybe we can try to appept something say 1000 times, and shutdown only after all-error?

reshke (approved):

PR #2825, opened 2026-08-03
Add MOVE KEY RANGE ... META ONLY command

EinKrebs:


reshke (approved):

Denchick:
https://docs.pg-sharding.tech/sharding/console/sql_commands#move-key-range

Is it a good idea to update the warning?

Could you add new syntax to the docs please?

PR #2829, opened 2026-08-03
Skip move task status `LOCKED` in QDB

diPhantxm:


test/feature/features/redistribute_retry.feature
```diff
@@ -813,7 +813,7 @@ Feature: Redistribution retries test
     key range is locked
     """
 
-Scenario: redistribute is retryable after fail to update KeyRangeMove to MoveKeyRangeLocked in QDB
```
EinKrebs, line 816: This scenario becomes obsolete and should be removed

EinKrebs (changes requested):

Denchick:
@diPhantxm could you rebase it so we can merge it?

PR #2830, opened 2026-08-03
Drop key range move status from QDB

diPhantxm:


qdb/etcdqdb.go
```diff
@@ -2273,7 +2273,11 @@ func (q *EtcdQDB) DropMoveTask(ctx context.Context, id string) error {
 		return err
 	}
 
-	_, err = q.cli.Txn(ctx).Then(clientv3.OpDelete(moveTaskNodePath(id)), clientv3.OpDelete(moveTaskByGroupNodePath(task.TaskGroupID))).Commit()
+	_, err = q.cli.Txn(ctx).Then(
+		clientv3.OpDelete(moveTaskNodePath(id)),
+		clientv3.OpDelete(moveTaskByGroupNodePath(task.TaskGroupID)),
+		clientv3.OpDelete(keyRangeMovesNodePath(id)),
```
EinKrebs: key range move has its own id, deleting it by move task id is ineffectual

EinKrebs (changes requested):

qdb/etcdqdb.go
diPhantxm: I found it on some cluster. Here krmove id is the same as moveId. Not sure now, but it seems we use moveTaskId according to the code. Or is it some other move_id and not moveTaskId?
```
/krmoves/12bb452b-ccac-46b8-9b28-ecdd3a00ebd5
{"move_id":"12bb452b-ccac-46b8-9b28-ecdd3a00ebd5", ... }
```

EinKrebs: it's key '/krmoves/move_id', value is *qdb.MoveKeyRange, move_id is written there as well. MoveId is actually different from move task id

protos/tasks.proto
```diff
@@ -117,6 +122,7 @@ service MoveTasksService {
     option deprecated = true;
   }
   rpc DropMoveTask(MoveTaskSelector) returns (google.protobuf.Empty) {}
```
EinKrebs, line 124: Consider deprecating this rpc

EinKrebs (approved):

protos/tasks.proto
reshke: separate commit for this

reshke (approved):

PR #2868, opened 2026-08-09
Beautify net coding after 0c5d55, also add some tests

reshke:
Golang runtime retries most of accept possible failure reasons, see 
https://github.com/golang/go/blob/25de5ebdcd710cd9dcb50ec6b03026fe3bc2a336/src/internal/poll/fd_unix.go#L608-L628. However, some errors are still not handled. go/http also has this kind of guard, see https://github.com/golang/go/blob/25de5ebdcd710cd9dcb50ec6b03026fe3bc2a336/src/net/http/server.go#L3595-L3611, so rework code committed at 0c5d558338f8be012c8488eadd96c5755b649e3a and add some simple tests

accept(2) explicitly recommends something like that at https://man7.org/linux/man-pages/man2/accept.2.html#RETURN_VALUE


rkhapov (approved):

diPhantxm (approved):

reshke:
hold

.github/actions/spelling/expect.txt
```diff
@@ -670,3 +670,12 @@ zerolog
 zjoin
 zzz
 zid
+ENONET
```
reshke, line 673: hold

PR #2897, opened 2026-08-17
Fix pool_discard and pool_rollback config placement across all router.yaml files

Denchick:
….yaml files


Denchick:
Screenshot 2026-08-18 at 00 14 59

PR #2902, opened 2026-08-19
Small improvements and beautifications in Adapter

diPhantxm:


pkg/coord/adapter.go
```diff
@@ -468,24 +387,13 @@ func (a *Adapter) Unite(ctx context.Context, unite *kr.UniteKeyRange) error {
...
-	}
-
-	return spqrerror.Newf(spqrerror.SPQR_KEYRANGE_ERROR, "key range with id %s not found", move.KeyRangeID)
+	c := proto.NewKeyRangeServiceClient(a.conn)
+	_, err := c.MoveKeyRange(ctx, &proto.MoveKeyRangeRequest{
+		Id:        move.KeyRangeID,
+		ToShardId: move.ShardID,
+		MetaOnly:  move.MetaOnly,
+	})
+	return spqrerror.CleanGrpcError(err)
```
reshke, line 396: Yep, we can avoid listing key ranges cause it's remote side duty to check for key range existence. 

But!

this dummy key range check would save us a little network time in case of missing key range. The general rule of thumb here: try to avoid coordinator interaction in cases it would be guaranteed erroneous.

But! This is all just unneeded for now, or should be separate PR.

reshke (approved):

PR #2936, opened 2026-08-26
spqradmin prototypes

Denchick:


Denchick:
https://github.com/denchick/spqradmin

PR #2947, opened 2026-08-31
Test SPQR against multiple Odyssey versions

Denchick:


EinKrebs:
I suggest to use separate compose file to run all of the feature tests, but for routers there to work through odyssey.
Also, there's no executions of odyssey feature tests in CI

PR #2954, opened 2026-09-02
add client and server TLS configuration for both router and coordinator

Denchick:


diPhantxm (approved):

cmd/router/main.go
```diff
@@ -298,7 +307,10 @@ var runCmd = &cobra.Command{
 			return fmt.Errorf("router failed to start: %w", err)
 		}
 
-		app := app.NewApp(router)
+		routerApp, err := app.NewApp(router)
```
reshke: why rename

docs/configuration/coordinator.mdx
```diff
@@ -40,6 +42,30 @@ Refer to the [pkg/config/coordinator.go](https://github.com/pg-sharding/spqr/blo
 | `use_spqrguard`            | Maintains [`spqrguard`](/deployment/spqrguard) metadata for key ranges and distributed and reference tables during topology changes and data moves. | `true`, `false` | `false` |
 | `forbid_direct_shard_queries` | Enables global `spqrguard` protection against direct writes to distributed and reference tables. | `true`, `false` | `false` |
 
+## gRPC TLS Settings
```
reshke, line 47: tls setup should be in separate file, so we can include from here and balancer/etc

reshke (approved):

PR #2957, opened 2026-09-03
Automatic rebootstrap of routers

EinKrebs:
* New RouterService RPCs are added: GetMetadataHash & Rebootstrap. GetMetadataHash computes hash of all distributions, relations & key ranges. Rebootstrap is router-specific RPC that works exactly as ALTER SYSTEM REBOOTSTRAP command.
* Add `watch_routers_qdb` setting to coordinator. When applied, coordinator compares contents of all routers' QDBs to its own. If contents differ, coordinator performs router re-bootstrap via gRPC.


reshke (approved):

Denchick (changes requested):
New setting but no docs. Could you add it please?

PR #2986, opened 2026-09-08
Escape shard id in spqr-monitor queries through router

EinKrebs:


reshke (approved):
not very safe but ok for now

PR #3014, opened 2026-09-18
Document SPQR virtual functions

### LINKED ISSUE #1918: Add documentation for __spqr__ special functions
SPQR provides several special functions prefixed with __spqr__, but these functions are not currently documented in the official documentation.
- __spqr__is_ready,
- __spqr__shards
- __spqr__show('key_ranges')

krylosov-aa:
Expand the virtual functions page with `__spqr__is_ready`, `__spqr__shards`, `__spqr__show` and `__spqr__host_status`

Fixes #1918


docs/configuration/virtual_function.mdx
```diff
@@ -1,9 +1,99 @@
...
+| `lower_bound` | Lower bound; multiple components are comma-separated. |
+| `locked` | Lock status: `'true'` or `'false'`. |
+
+Use the function in `FROM` to select columns and filter rows:
+
+```sql
+SELECT key_range_id, shard_id FROM __spqr__show('key_ranges')
+WHERE shard_id = 'sh1';
+
+SELECT * FROM __spqr__show('key_ranges') WHERE locked = 'false';
```
krylosov-aa, line 85: I didn't include an `ORDER BY` example, but it may be useful to add one. Without knowing about `__spqr__allow_postprocessing`, it's not obvious why sorting doesn't work.

docs/configuration/virtual_function.mdx
```diff
@@ -1,9 +1,99 @@
 ---
-Virtual functions
+title: 'Virtual functions'
+description: 'Check router readiness and inspect SPQR metadata through SQL'
 ---
 
+Virtual functions let you check router status and query SPQR metadata from an
+application database connection.
```
EinKrebs:
```suggestion
application database connection. It also allows to run some administrative commands
```

docs/configuration/virtual_function.mdx
EinKrebs: The `__spqr__allow_postprocessing` GUC and router:query_router:allow_postprocessing parameter controls whether router can perform basic aggregation and sorting of query results. When turned off, router simply returns unprocessed rows. It's turned off by default

EinKrebs: Can you please also include ORDER BY examples, with the mention of `allow_postprocessing`

EinKrebs (changes requested):

krylosov-aa:
Done

krylosov-aa:
I didn't include a direct example because it seemed like it would look silly. So, I simply noted that using `ORDER BY` requires enabling a specific setting and that there are three ways to do it.

EinKrebs (approved):

Issue #2772, opened 2026-07-24
Data race / concurrent map access on QueryStateExecutorImpl.localConn during relay close


graysoncooper:
**Describe the bug**
`QueryStateExecutorImpl.localConn` (router/relay/executor.go:57) is a plain `map[string]shard.ShardHostInstance` with no mutex or ownership barrier, yet it is read and written from lifecycle paths that can overlap:

- `Close()` iterates it and returns each connection to the pool (executor.go:63-67)
- `Discard()` deletes from it (executor.go:70-72)
- `ConnectionWithTSA()` reads it (executor.go:95)
- `CleanupConnection()` inserts into it (executor.go:128-130)

`RelayStateImpl.Close()` calls `QueryExecutor().Close()` during relay shutdown (router/relay/relay.go:234-236), which runs when a client disconnects. If a client disconnects while a routed request is still allocating, reading, or cleaning up a pinned connection, `Close()` races the in-flight access on the same map. Go maps are not safe for concurrent use, so this can raise a fatal `concurrent map ...` runtime panic that takes down the router, or return a backend connection to the pool while it is still in use.

Introduced by #2768 (connection pinning), which added `localConn` and its access paths.

**To Reproduce**
1. Enable session connection pinning (`session_connections_pin`).
2. Start a routed request that obtains or cleans up a pinned shard connection.
3. While the request is still allocating/reading/inserting into the cache, disconnect the client so `RelayStateImpl.Close()` runs.
4. Run under the race detector (`-race`) or repeat under concurrency and watch for `fatal error: concurrent map read and map write` (or `... iteration and map write`), or a backend returned to the pool mid-request.

**Expected behavior**
Access to `localConn` should be synchronized (a mutex or single-owner handoff), and `Close()` should coordinate with in-flight cache operations so a disconnect during an active request cannot panic or release an in-use connection.

**Additional context**
Permalinks (HEAD 18e662d):
- decl: https://github.com/pg-sharding/spqr/blob/18e662d0373f242a9d0f282a6425a426f6155b7a/router/relay/executor.go#L57
- Close: https://github.com/pg-sharding/spqr/blob/18e662d0373f242a9d0f282a6425a426f6155b7a/router/relay/executor.go#L63-L67
- Discard: https://github.com/pg-sharding/spqr/blob/18e662d0373f242a9d0f282a6425a426f6155b7a/router/relay/executor.go#L70-L72
[...]

PR #3015, opened 2026-09-19
Fix pinned connection ownership and cleanup

krylosov-aa:
A pinned connection could remain in the local cache after being returned to the pool. Closing the client could then return the same connection again and crash the router.

Remove cached connections when they are handed to a server, and cache them again during cleanup if pinning is still enabled. Release connections when cleanup or starting a transaction on another shard fails. Keep the server attached during client shutdown so the relay can release its connections.

Related to #2772.

I could not reproduce the concurrent map access reported in #2772. In the inspected code, request processing and the deferred `relay.Close()` run in the same goroutine. Disconnecting the client does not start a separate executor cleanup routine.

The investigation did reveal related connection lifecycle problems: returning a pinned connection to the pool twice, leaking connections on cleanup or transaction expansion errors, and losing an acquired connection during shutdown. These are addressed by this change.


krylosov-aa:
tsa_5node is flaky. Fixed it in #3016

PR #3016, opened 2026-09-19
Stabilize `tsa_5node` TSA cache assertions

krylosov-aa:
`tsa_5node` sometimes failed because test expected a `read-write` cache entry for every replica. However, the router checks hosts in a partly random order and stops when it finds the primary. Some replicas may never be checked, so their cache entries may be missing even when routing works correctly.

The test now ignores these optional entries and checks `read-write` entries only for the primary and the local replica, which is checked first because of its higher priority. This keeps the check that a replica can be alive but unsuitable for `read-write`. The existing checks for replica routing and fallback to the primary when all replicas are unavailable remain unchanged.

This showed up on #3015


EinKrebs (changes requested):
It will be more suitable to write a helper virtual function, that will preheat caches for tsa, rather than performing less rigorous check

krylosov-aa:
@EinKrebs

I corrected the PR as requested

